PR #7, opened 2019-05-23
travis: validate each package.json

glensc:
validates each package `composer.json`:

```
➔ for COMPOSER in packages/*/composer.json; do COMPOSER=$COMPOSER composer.phar validate; done
packages/zend-acl/composer.json is valid
packages/zend-amf/composer.json is valid
packages/zend-application/composer.json is valid
packages/zend-auth/composer.json is valid
packages/zend-barcode/composer.json is valid
packages/zend-cache/composer.json is valid
packages/zend-captcha/composer.json is valid
^C
```

https://getcomposer.org/doc/03-cli.md#composer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,4 +42,9 @@ before_script:
   - cp ./tests/TestConfiguration.travis.php ./tests/TestConfiguration.php
 
 script:
+  # validate that every composer.json is valid
+  # https://github.com/zf1s/zf1/pull/6#issuecomment-495397170
+  - for json in packages/*/composer.json; do COMPOSER=$json composer validate; done
   - ./vendor/bin/phpunit
+
+# EOF


.travis.yml
```diff
@@ -42,4 +42,9 @@ before_script:
   - cp ./tests/TestConfiguration.travis.php ./tests/TestConfiguration.php
 
 script:
+  # validate that every composer.json is valid
+  # https://github.com/zf1s/zf1/pull/6#issuecomment-495397170
+  - for json in packages/*/composer.json; do COMPOSER=$json composer validate; done
   - ./vendor/bin/phpunit
+
+# vim:ts=2:sw=2:et
```
partikus: I would drop this out in favor of [editorconfig](https://editorconfig.org/)

partikus: @glensc here you go: https://github.com/zf1s/zf1/blob/master/.editorconfig

glensc: you know, the vim implementation is the poorest of editorconfig. it requires you to have specific build of vim with python enabled, and bunch of other random bundle/vundle/blurb.

and while i mostly edit yaml files on servers where is just plain vim, i find the modeline just do the thing. especially the default in vim is tab which is invalid for yaml.

the situation is better with neovim: it has better python support, no need to recompile, just install addon package. but the editor config integration still needs to be loaded external.

however, i'll remove from this project after this rant :D

glensc: the other reason I leave some comment at the end of file, is that `insert_final_newline` setting which each editor handles differently. and i use those conflicting editors daily. luckily with last line being something irrelevant I can decide not to change that chunk with `git add -p`

glensc: https://github.com/zf1s/zf1/pull/9

partikus: ok, I see but I cannot help you with VIM approach too much ... 
I would adjust `.editorconfig` file that is a well-known standard to adjust the editor's configuration
regarding the VIM config, I would avoid putting too many users dependant stuff to the main repository ... maybe there is another way that solves the issue?

Issue #11, opened 2019-07-09
[zend-file-transfer]: unable to load adapters


glensc:
given code:
```php
<?php
require __DIR__ . '/vendor/autoload.php';
$file = new Zend_File_Transfer();
```

with `zf1s/zend-file-transfer` installed, i get error:
```
PHP Fatal error:  Uncaught Error: Class 'Zend_Loader' not found in vendor/zf1s/zend-file-transfer/library/Zend/File/Transfer.php:67
Stack trace:
#0 vendor/zf1s/zend-file-transfer/library/Zend/File/Transfer.php(54): Zend_File_Transfer->setAdapter('Http', false, Array)
#1 test.php(5): Zend_File_Transfer->__construct()
#2 {main}
  thrown in vendor/zf1s/zend-file-transfer/library/Zend/File/Transfer.php on line 67
```

loading `"zf1s/zend-loader"`, the error changes to:

```
PHP Warning:  include_once(Http.php): failed to open stream: No such file or directory in vendor/zf1s/zend-loader/library/Zend/Loader.php on line 134
PHP Stack trace:
PHP   1. {main}() test.php:0
PHP   2. Zend_File_Transfer->__construct() test.php:5
PHP   3. Zend_File_Transfer->setAdapter() vendor/zf1s/zend-file-transfer/library/Zend/File/Transfer.php:54
PHP   4. Zend_Loader::loadClass() vendor/zf1s/zend-file-transfer/library/Zend/File/Transfer.php:72
PHP   5. Zend_Loader::loadFile() vendor/zf1s/zend-loader/library/Zend/Loader.php:82
```

I've created gist to play around with:
https://gist.github.com/glensc/ce6a301ddc60a0d1dba8878ae32709c3

with git tags `v1` and `v2` respectively for the tests.

glensc:
some thoughts: as constructor calls `setAdapter` unconditionally:
- https://github.com/zf1s/zf1/blob/1.13.0/packages/zend-file-transfer/library/Zend/File/Transfer.php#L54

which calls `Zend_Loader` unconditionally:
- https://github.com/zf1s/zf1/blob/1.13.0/packages/zend-file-transfer/library/Zend/File/Transfer.php#L67

`zend-loader` should be added as package dependency.

glensc:
the current code ends up calling `Zend_Loader::loadClass(Http)` because `$adapter = 'Zend_File_Transfer_Adapter_' . ucfirst($adapter);` is never assigned.

```php
[...]

PR #12, opened 2019-07-09
Zend_File_Transfer: composer safe setAdapter method

glensc:
Fixes: #11

### DIFF
--- a/packages/zend-file-transfer/library/Zend/File/Transfer.php
+++ b/packages/zend-file-transfer/library/Zend/File/Transfer.php
@@ -64,11 +64,9 @@ class Zend_File_Transfer
      */
     public function setAdapter($adapter, $direction = false, $options = array())
     {
-        if (Zend_Loader::isReadable('Zend/File/Transfer/Adapter/' . ucfirst($adapter). '.php')) {
+        if (class_exists('Zend_File_Transfer_Adapter_' . ucfirst($adapter))) {
             $adapter = 'Zend_File_Transfer_Adapter_' . ucfirst($adapter);
-        }
-
-        if (!class_exists($adapter)) {
+        } elseif (!class_exists($adapter)) {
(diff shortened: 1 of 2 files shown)


packages/zend-file-transfer/library/Zend/File/Transfer.php
```diff
@@ -64,18 +63,15 @@ public function __construct($adapter = 'Http', $direction = false, $options = ar
...
     public function setAdapter($adapter, $direction = false, $options = array())
     {
-        if (Zend_Loader::isReadable('Zend/File/Transfer/Adapter/' . ucfirst($adapter). '.php')) {
-            $adapter = 'Zend_File_Transfer_Adapter_' . ucfirst($adapter);
-        }
-
-        if (!class_exists($adapter)) {
-            Zend_Loader::loadClass($adapter);
-        }
+	    if (class_exists('Zend_File_Transfer_Adapter_' . ucfirst($adapter))) {
```
falkenhawk: could you adjust the indentation here please?

packages/zend-file-transfer/library/Zend/File/Transfer.php
```diff
@@ -22,7 +22,6 @@
 /**
  * @see Zend_Loader
  */
-// require_once 'Zend/Loader.php';
```
falkenhawk: do not remove commented out `require_once`s, please, as they are simply left there to be in all other occurrences - I would not do any code administrative/beautifying/cleaning tasks in the whole framework and keep the code changes only to the really necessary ones.

packages/zend-file-transfer/library/Zend/File/Transfer.php
glensc: powers of web edit. looks like have to check this out locally. thanks for the reivew.

glensc:
cleaned up now

PR #14, opened 2019-11-29
[PHP 7.4 Compat] Remove overloaded offsetExists()

Dave13h:
This blows up on PHP 7.4 because of the change to `array_key_exists()` which now strictly expects an `array`.
https://www.php.net/manual/en/migration74.deprecated.php#migration74.deprecated.core.array-key-exists-objects

This method looks like it was only added because of this bug https://bugs.php.net/bug.php?id=40442 which is exclusive to PHP 5.2.1 and was fixed in 5.2.2 - https://www.php.net/ChangeLog-5.php#5.2.2

(copied over from https://github.com/zf1s/zend-registry/pull/1 )

### DIFF
--- a/packages/zend-registry/library/Zend/Registry.php
+++ b/packages/zend-registry/library/Zend/Registry.php
@@ -194,16 +194,4 @@ class Zend_Registry extends ArrayObject
     {
         parent::__construct($array, $flags);
     }
-
-    /**
-     * @param string $index
-     * @returns mixed
-     *
-     * Workaround for http://bugs.php.net/bug.php?id=40442 (ZF-960).
-     */
-    public function offsetExists($index)
-    {
-        return array_key_exists($index, $this);
-    }
-
 }


falkenhawk:
Whoops, it looks like I have to adjust travis config...

falkenhawk:
Since there are a lot more compatibility issues with php7.4 (https://travis-ci.com/zf1s/zf1/jobs/261619983#L350) I am going to convert it into a bigger PR to get rid of them all.

holtkamp:
>Since there are a lot more compatibility issues with php7.4

Maybe [rector](https://github.com/rectorphp/rector) can help with making ZF1 compatible with PHP 7.4, as mentioned in this tweet https://twitter.com/ArkadiuszKondas/status/1200103030823280640 🤓 

Please let me know in case help is needed.

Megatherium:
Hi,
I went through all tests yesterday and they all pass now for 7.4 bar one but that seems to be a problem with the sqlite driver and is probably something not as trivial to fix: https://travis-ci.com/Megatherium/zf1/jobs/263793214

I've done a bit more than just the functional stuff (updated docblocks, used more proper assertions, simplified a bit) for every file that I had to touch. Would you like this as a PR? I could also chop it up and do one for each module if that's more to your liking.

Cheers

falkenhawk:
@Megatherium great job! Thanks for notifying us. Of course I'd love a PR - since this is a monorepo, one is fine, no need to split it into separate ones for each component.
Did you use a tool for that, e.g. rector as suggested here? Or a different one? I only hope you didn't do all of those changes by hand... I'd only suggest resorting from changing too much stuff especially reformatting docblocks (except where param or return types are fixed - those are welcome changes of course) or other formatting-only changes, in order not to change the original code too much and only when really necessary. But we could continue the discussion in your PR when you set it up.
Thanks again! The open source community is really amazing...

falkenhawk:
superseded by #16

PR #16, opened 2019-12-06
Make framework PHP 7.4 compatible

Megatherium:
There were a bunch a minor changes needed like trimming slashes and
octothorpes of strings before passing them into hexdec and removing all
array accesses via curly braces.

I have no idea how Zend_Crypt_Math::rand ever worked under unixlikes
because we expect it to return a number but it outputs directly from
/dev/urandom if it exists.

Serialization of ReflectionMethods is also illegal since 7.4

A lot of docblocks were updated, assertions changed to more obvious ones
and grimey things like dirname(__FILE__) were properly removed whenever
I encountered them for most files that I touched.


falkenhawk:
@Megatherium could you modify the travis config and replace `7.4snapshot` with `7.4`, and remove `7.4` from `allow_failures`, please?

packages/zend-crypt/library/Zend/Crypt/Math.php
```diff
@@ -49,7 +49,8 @@ public function rand($minimum, $maximum)
         if (file_exists('/dev/urandom')) {
             $frandom = fopen('/dev/urandom', 'r');
             if ($frandom !== false) {
-                return fread($frandom, strlen($maximum) - 1);
+                # I don't know how this could've ever worked if you're expecting an integer output
+                #return fread($frandom, strlen($maximum) - 1);
             }
         }
```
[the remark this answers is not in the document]
Megatherium: Removed the whole block. It almost seems as if somebody just copy/pasted from the randBytes function.

Megatherium:
I basically just used PHPStorm with some plugins and spammed a lot of Alt+Enter for the docblocks. Most of the changes to the code (that weren't necessitated by failing tests) were also suggested by the IDE but I was rather conservative in what I actually changed especially when it came to changing == to ===.

Megatherium:
Generalized the expected exception text so we don't need any version compares. Holler if you need anything else done.

falkenhawk:
Released version 1.13.1 of all packages.

packages/zend-service-rackspace/library/Zend/Service/Rackspace/Servers/Server.php
```diff
@@ -104,11 +104,11 @@ public function __construct($service, $data)
             // require_once 'Zend/Service/Rackspace/Servers/Exception.php';
             throw new Zend_Service_Rackspace_Servers_Exception(self::ERROR_PARAM_CONSTRUCT);
         }
-        if (!array_key_exists('name', $data)) {
+        if (!isset($data['name'])) {
```
holtkamp: Note that `array_key_exists` will return `true` of an array value is `NULL` while `isset` will return `false`. 

So this is a functional change....

https://3v4l.org/6llcH

falkenhawk: You are right @holtkamp ... @Megatherium could you elaborate on this change, please? Did it even make sense to allow `NULL` values in `id` and `name` before? But I tend to agree this should not go like that into a patch release...

Megatherium: I have to admit that I was overeager and ignorant of the minute difference.

PR #20, opened 2020-01-11
fixed Incompatiblity with standard Zend MVC, Zend module MVC

saru2048:
Hello.
Thanks for publishing this repository.
I found two bugs and fixed them, so I create a pull request.
Please consider it.

 * A request url, controller does not exist, it not got a 404 error.
 * Default/controllers/IndexController load fails when default module prefix is disabled
     * classname is IndexController
     * $front->prefixDefaultModule(false)

Regards,


falkenhawk:
@saru2048 I have changed these `class_exists` calls in https://github.com/zf1s/zf1/commit/cadf3e058d6b9989c89d07ccbabe6fe020bcd9ad because of apparently some issues with composer autoloader, which I don't remember exactly anymore :(

Too bad it broke the thing for you. But I am afraid changing it back may break composer autoloading again.

Could you maybe add a test case to reproduce this, please?

falkenhawk:
Fixed in #116 (most probably.)

PR #21, opened 2020-05-27
Add Croatia to the European Union

dmnc:
https://europa.eu/european-union/about-eu/countries_en#tab-0-1


falkenhawk:
Thank you @dmnc . Could you fix the failing test, please?

Issue #25, opened 2020-08-26
Merge with zf1-future


glensc:
Merge efforts with shardj/zf1-future project:
- https://github.com/Shardj/zf1-future

Seems they have gone further original plan to support modern PHP: Minimum version required is PHP 7.4.

So the project goals are not the same, but there may be large overlapping efforts.

x-ref:
- https://github.com/Shardj/zf1-future/issues/96

glensc:
They were thinking restoring backward compat from 7.0:
- https://github.com/Shardj/zf1-future/issues/94

but seems now their minimum is 7.4 instead...

Shardj:
It should still be 7.0 minimum, I don't think 7.4 has ever been the minimum. If it was I'd be pretty screwed since all of my systems were on 7.2 until this morning

glensc:
# carry status: 2022-10-04

## already applied

- [x] https://github.com/Shardj/zf1-future/pull/167

## carried

- [x] https://github.com/zf1s/zf1/pull/95
- [x] https://github.com/zf1s/zf1/pull/96
- [x] https://github.com/zf1s/zf1/pull/97
- [x] https://github.com/zf1s/zf1/pull/98
- [x] https://github.com/zf1s/zf1/pull/99
- [x] https://github.com/zf1s/zf1/pull/100
- [x] https://github.com/zf1s/zf1/pull/102
- [x] https://github.com/zf1s/zf1/pull/103
- [x] https://github.com/zf1s/zf1/pull/104
- [x] https://github.com/zf1s/zf1/pull/105
- [x] https://github.com/zf1s/zf1/pull/131

## more work needed
- [x] https://github.com/Shardj/zf1-future/pull/75 - needs more work, keep catching old exception

## irrelevant
- https://github.com/Shardj/zf1-future/pull/145
- https://github.com/Shardj/zf1-future/pull/143

Shardj:
Lots of respect for the amount of work you do for zf1s and the help you provide to zf1-future @glensc

glensc:
Some unknown files committed in (origin and author unclear):
- https://github.com/Shardj/zf1-future/pull/38#issuecomment-1274704685

glensc:
@falkenhawk I was trying to cherry-pick some PHP 8.1 changes from zf1-future, and it seemed need to cherry-pick some earlier changes, so ended up picking:
- https://github.com/Shardj/zf1-future/pull/192

[...]

PR #27, opened 2020-10-01
Zend-Xml php8 compatibility

leendt:
In Zend_Xml_Security: `libxml_disable_entity_loader()` is no longer executed when running PHP 8 or higher. It's still executed (and necessary!) on prior versions.

This restores the ability to load an XML configuration file using Zend_Config_Xml when running PHP 8. Now the deprecated warning will cause it to not load anything and throw an exception: `Zend_Config_Exception: Function libxml_disable_entity_loader() is deprecated`.

I've also renamed `Zend_Xml_Security::isPhpFpm()`  to `Zend_Xml_Security::isVulnerablePhpFpm()` in the second commit because that name better represents what the method does in my opinion.

### DIFF
--- a/packages/zend-xml/library/Zend/Xml/Security.php
+++ b/packages/zend-xml/library/Zend/Xml/Security.php
@@ -46,6 +46,19 @@ class Zend_Xml_Security
     }
 
     /**
+     * Wrapper for libxml_disable_entity_loader which is deprecated in PHP 8. libxml 2.9.0 disables external entity
+     * loading by default. PHP 8.0 has deprecated libxml_disable_entity_loader() method and it requires libxml 2.9.0+.
+     * @see https://github.com/php/php-src/pull/5867
+     *
+     * @param bool $disable
+     * @return bool the previous value (true = disabled)
+     */
+    private static function disableEntityLoader($disable = true)
+    {
+        return LIBXML_VERSION < 20900 ? libxml_disable_entity_loader($disable) : true;
+    }
+
+    /**
      * @param integer $errno
      * @param string $errstr
      * @param string $errfile
@@ -83,7 +96,7 @@ class Zend_Xml_Security
         }
 
         if (!self::isPhpFpm()) {
-            $loadEntities = libxml_disable_entity_loader(true);
+            $loadEntities = self::disableEntityLoader(true);
             $useInternalXmlErrors = libxml_use_internal_errors(true);
         }
 
@@ -97,7 +110,7 @@ class Zend_Xml_Security
         if (!$result) {
             // Entity load to previous setting
             if (!self::isPhpFpm()) {
-                libxml_disable_entity_loader($loadEntities);
+                self::disableEntityLoader($loadEntities);
                 libxml_use_internal_errors($useInternalXmlErrors);
             }
             return false;
@@ -117,7 +130,7 @@ class Zend_Xml_Security
 
         // Entity load to previous setting
         if (!self::isPhpFpm()) {
-            libxml_disable_entity_loader($loadEntities);
+            self::disableEntityLoader($loadEntities);
             libxml_use_internal_errors($useInternalXmlErrors);
         }
 


leendt:
This is my first PR on Github. If it's lacking or if you have any advice for improvement please let me know.

packages/zend-xml/library/Zend/Xml/Security.php
```diff
@@ -45,6 +45,19 @@ protected static function heuristicScan($xml)
         }
     }
 
+    /**
+     * Wrapper for libxml_disable_entity_loader which is deprecated in PHP 8. libxml 2.9.0 disables external entity
+     * loading by default. PHP 8.0 and later has the option to enable it at runtime removed and libxml 2.9.0+ enforced.
```
glensc, line 50: I found your sentence difficult to read due to weird language construct. perhaps this one is better?

```
     * loading by default. PHP 8.0 has deprecated libxml_disable_entity_loader() method and it requires libxml 2.9.0+.
```

glensc:
I don't think this project should be doing improvements like renaming things. makes it difficult to keep in sync with other similar projects: https://github.com/zf1s/zf1/issues/25. but I guess it's up. to maintainers to decide.

glensc:
https://github.com/Shardj/zf1-future porting notice:
- not covered there: https://github.com/Shardj/zf1-future/commits/release-1.18.6/library/Zend/Xml/Security.php

packages/zend-xml/library/Zend/Xml/Security.php
```diff
@@ -164,7 +177,7 @@ public static function scanFile($file, DOMDocument $dom = null)
      *
      * @return boolean
      */
-    public static function isPhpFpm()
+    public static function isVulnerablePhpFpm()
```
falkenhawk, line 180: Although I agree with @leendt that the `isVulnerablePhpFpm` name would suit it better, I second @glensc we should not introduce unnecessary changes (and definitely not breaking ones).

To avoid breaking changes, we should still have `isPhpFpm()` which would be an alias to `isVulnerablePhpFpm()`, but still `isPhpFpm()` would keep its ambiguous meaning...

I'd say better to keep it as it was. I can see in the method's docblock that the purpose of it is explained, and I guess from it that the original zf1 maintainers first introduced the method as a workaround, and then they didn't care to rename it when php-fpm had been eventually fixed.

Now I wouldn't like to start any active development on this repo (like renaming stuff, adding new features, tidying up the code to make it up to current standards etc.) since the purpose of it is to just help people "stuck" with zf1 (where it's simply economically impossible to update their projects to different, more modern frameworks) to keep their systems running also on current php versions, **while also** fixing evident issues and bugs whenever possible, but always with BC intact. After all, it's simply not encouraged to start a new project based on ancient zf1, when there are so many, much better, modern alternatives...

packages/zend-xml/library/Zend/Xml/Security.php
```diff
@@ -19,7 +19,7 @@
  * @version    $Id$
  */
 
- 
+
```
falkenhawk, line 22: unnecessary blank line added

falkenhawk (commented):
Thank you very much @leendt for your contribution!

I guess php8 compatibility will be a much bigger topic, have you noticed any other issues already?

packages/zend-xml/library/Zend/Xml/Security.php
glensc: the link in same docblock said it's deprecated:
- https://github.com/php/php-src/pull/5867

glensc: but the PR suggest checking for libxml instead:

```
if (LIBXML_VERSION < 20900) {
    libxml_disable_entity_loader();
}
```

- https://github.com/php/php-src/pull/5867#issuecomment-667941848

packages/zend-xml/library/Zend/Xml/Security.php
leendt: Actually the line was already there, an unnecessary white space was removed.

falkenhawk: oh my bad 🙈 anyway, let's avoid unnecessary changes in code formatting

packages/zend-xml/library/Zend/Xml/Security.php
leendt: PHP 8.0rc1 gives a deprecated warning indeed. That's what made me look into it.

I agree with the comment. English is not my first language.

I've noticed the libxml version check and used that at first.
However, after reading up I assumed that one can still execute `libxml_disable_entity_loader(false)` on PHP versions < 8 with LibXML version >= 2.9.0 to enable entity auto-loading globally. And that this could still be problematic.

While trying to look up documentation to show it to you I realized I lack(ed) proper understanding of what the function/workaround does! I'll dive into it to get a better understanding and reply again (and of course reconsider).

Thank you (all) for taking the time for an in-depth review.

packages/zend-xml/library/Zend/Xml/Security.php
leendt: I agree and will revert that commit.

Active development is not my goal either.

[some bacground]
We're stuck with a behemoth of an (internal) application deeply coupled with (now stock) ZF1 1.12. It'll have to be refactored, a full rewrite or keeping up two applications is a path I want to avoid if possible due to cost and the risk of repeating old 'mistakes'.

I was very happy to find this project. Not only are most PHP 7 deprecateds/warnings and some bugs fixed, the components are also split up as far as possible. After registering ZendX with the Zend autoloader (and removing the require_onces) it was a drop in replacement and I was very surprised it also ran without major issues in PHP 8.

Using this could enable us to to take more time to 'unwire' our own code/logic from Zend1 and replace components gradually while making use of everything PHP 7 and soon 8 will offer.
[/some background]

packages/zend-xml/library/Zend/Xml/Security.php
glensc: this is how I understand the problem:

- the whole problem is around disabling XXE attack vectors
- XXE can be disabled with libxml_disable_entity_loader(false), but it disables some more functionality, that could be useful (read discussion in linked php-src PR)
- libxml 2.9.0 disables XXE by default, so libxml_disable_entity_loader(false) call is not needed
- PHP 8 requires libxml 2.9.0 to be used (won't compile with older version)
- you can compile PHP < 8 with libxml 2.9.0
- PHP 8 adds deprecation to calls to libxml_disable_entity_loader(), altho not clear to me at this point is it noop call (doing nothing) or it still enables/disables something
- zend-xml calls libxml_disable_entity_loader() conditionally, so it must be okay to call or omit the call.

so, summarize that all: use libxml version check for all PHP versions

leendt: You are right. According to https://github.com/php/php-src/pull/5867#issuecomment-664213053 calling `libxml_disable_entity_loader(false)` by itself doesn't do anything when compiled with libxml => 2.9.0, no matter the PHP version.

So the version check appears to be safe and the best option. I've verified this by testing with PHP 5.3.1, Libxml 2.7.3 and 5.6, 7.4 and 8 with 2.9.0+ using the script found here: https://gist.github.com/lukaskuzmiak/c8306a5af855c6faaaee . Only passing 'LIBXML_NOENT' will enable loading the external file on PHP with 2.9.0+.

To summarize:
- Revert the commit with the protected method name change (to keep changes to a minimum  for future syncing).
- Revert the removal of the unnecessary space to keep file formatting changes to a minimum.
- Replace the PHP version check with a libxml version check.
- Update the comment for the wrapper method to be less weird.

Thanks again for all the feedback. Since this is a first for me I'm unsure about the correct way to present the changes. Should I close this PR and start a new one? Or commit reverts and then the (hopefully) proper fix?

glensc: I personally would rewrite the branch as intended and then force push new changes.

falkenhawk: @leendt you may also just push another commit(s) with changes, I'll eventually squash on merge to one commit anyway.

leendt: Thanks for the compliments! I'm glad it shows that I'm putting quite some effort in understanding and getting it right. The in depth reviews were also quite motivating.

As for the old revisions, they'd become dangling commits after a rebase/rewrite, right? That's nice, still accessible yet automatically purged after a while.

I've reverted and committed the revert and the changes and pushed them. After that I've squashed them and force pushed the resulting commit because I was curious as to what it would do in this conversation view.

I was afraid GitHub might remove or move the conversation with a rebase + force push due to the branch rewriting. Glad to see GitHub handles it chronologically.

falkenhawk (approved):

packages/zend-xml/library/Zend/Xml/Security.php
glensc: I personally avoid referencing to commits that are not merged, especially the ones that do not exist (will not exist). you currently pushed commit (6d970a786378025680619649494699ec788d6bf4) message refers to rewritten commit, which will not exist once merged to master. i.e from local git clones. altho it will keep existing in GitHub UI.

perhaps you want to rewrite whole commit message? that you had commit that you removed is not interesting (useful) after this is merged to the default branch.

leendt: Good point. I've reworded it.

packages/zend-xml/library/Zend/Xml/Security.php
```diff
@@ -45,6 +45,19 @@ protected static function heuristicScan($xml)
...
+     * Wrapper for libxml_disable_entity_loader which is deprecated in PHP 8. libxml 2.9.0 disables external entity
+     * loading by default. PHP 8.0 has deprecated libxml_disable_entity_loader() method and it requires libxml 2.9.0+.
+     * @see https://github.com/php/php-src/pull/5867
+     *
+     * @param bool $disable
+     * @return bool the previous value (true = disabled)
+     */
+    protected static function disableEntityLoader($disable = true)
+    {
+        return LIBXML_VERSION < 20900 ? libxml_disable_entity_loader($disable) : true;
```
glensc, line 58: 1. are you sure you want to always return true, even if you intend to disable? seems you want to return `!$disable`, so the api would behave like the change "worked".
3. make method private, makes maintenance easier if things not exported too much
2. the return value is not used at all. so make method void?

glensc: EDIT: the return value **is** used, mislooked the usage. and as the method is private does not matter what is returned in the "no-op" mode.

leendt: 1. There's something to be said for that, the toggling still works even on PHP8 with the deprecated warning. As you said it doesn't matter here but I might as well do it 'right'.
2.  I went with the standard in that file. Though I doubt there's code out there extending that Security class and wanting to use this method so private is also fine with me.

Should I make both changes?

glensc: make it only `private`.

leendt: Done.

glensc (approved):

packages/zend-xml/library/Zend/Xml/Security.php
```diff
@@ -83,7 +96,7 @@ public static function scan($xml, DOMDocument $dom = null)
         }
 
         if (!self::isPhpFpm()) {
-            $loadEntities = libxml_disable_entity_loader(true);
+            $loadEntities = self::disableEntityLoader(true);
```
partikus, line 99: @leendt is it necessary to pass `true` since it's a default value?

partikus: Besides that, it looks fine.

leendt: No it's not necessary. It was not necessary for the original PHP function either:
```php
/**
 * Disable the ability to load external entities
 * @link https://php.net/manual/en/function.libxml-disable-entity-loader.php
 * @param bool $disable [optional] <p>
 * Disable (<b>TRUE</b>) or enable (<b>FALSE</b>) libxml extensions (such as
 * ,
 * and ) to load external entities.
 * </p>
 * @return bool the previous value.
 * @since 5.2.11
 */
function libxml_disable_entity_loader ($disable = true) {}
```

The original Zend code probably passed it for readability/clarity.

I've copied the `libxml_disable_entity_loader` function signature for the same reason but I admit that's a weak one since it's a private method where the bool is always passed anyway. 
We can either remove the default value from the wrapper or stop passing a value.

Should I change it?

glensc: you need to pass the value when the restore happens for libxml < 2.9.0, I'll say leave as is.

Issue #28, opened 2020-10-07
Zend_Db_Adapter_Pdo_Pgsql is broken with Postgres 12


Megatherium:
https://github.com/zf1s/zf1/blob/430645172897d6b89d59540d16399eec198bddbf/packages/zend-db/library/Zend/Db/Adapter/Pdo/Pgsql.php#L159

[adsrc has been considered 'historical' since PG8](https://www.postgresql.org/docs/8.0/catalog-pg-attrdef.html) and has finally been removed with PG12. The line can easily be replaced with `pg_get_expr(d.adbin, d.adrelid) AS default_value,` I have tested this down to PG9.6 but from reading the documentation this should work even with <8.0

Will this suffce?

glensc:
If what you're saying, that the new proposed version works for PG < 8.0 and PG >= 8.0, I see no reason why not to make the proposed change.

glensc:
Solved the same way in Shardj/zf1-future (https://github.com/zf1s/zf1/issues/25):

- https://github.com/Shardj/zf1-future/issues/28
- https://github.com/Shardj/zf1-future/pull/31

so perhaps just cherry-pick the change.

glensc:
there: https://github.com/zf1s/zf1/pull/29

PR #29, opened 2020-10-07
Fix Zend_Db_Adapter_Pdo_Pgsql being broken on Postgres 12 that removed d.adsrc

glensc:
fixes https://github.com/zf1s/zf1/issues/28

cherry-picked from:
- https://github.com/Shardj/zf1-future/pull/31
- https://github.com/Shardj/zf1-future/issues/28


falkenhawk:
Does the change have any significant impact which should be mentioned as a BC break? (e.g. required postgresql version)

glensc:
@falkenhawk please see #28 discussion. seems to me it's not a BC break.

falkenhawk (approved):

Issue #31, opened 2020-10-19
iconv_substr behaves differently under PHP8


Megatherium:
While testing out software with PHP8 (8.0.0~rc1-6+ubuntu18.04.1+deb.sury.org+1) I got a TypeError from mktime during basic usage of Zend_Date.
```
<?php

require_once 'vendor/autoload.php';

$date = new Zend_Date('2020-01-01');
$date->set(0, Zend_Date::TIMES);
echo $date.PHP_EOL;
```
Resulted in this with PHP8
```
PHP Fatal error:  Uncaught TypeError: gmmktime(): Argument #2 ($minute) must be of type ?int, string given in /somepath/vendor/zf1s/zend-date/library/Zend/Date/DateObject.php:156
Stack trace:
#0 /somepath/vendor/zf1s/zend-date/library/Zend/Date/DateObject.php(156): gmmktime()
#1 /somepath/vendor/zf1s/zend-date/library/Zend/Date.php(2277): Zend_Date_DateObject->mktime()
#2 /somepath/vendor/zf1s/zend-date/library/Zend/Date.php(1078): Zend_Date->_calculate()
#3 /somepath/zd.php(7): Zend_Date->set()
#4 {main}
  thrown in /somepath/vendor/zf1s/zend-date/library/Zend/Date/DateObject.php on line 156
```

`mktime` isn't the culprit in this case though. It turned out that Zend_Locale_Format was acting differently:
```
<?php

require_once 'vendor/autoload.php';

var_dump(Zend_Locale_Format::getTime(0));
```
7.4:
```
array(5) {
  'date_format' =>
  string(9) "h:mm:ss a"
  'locale' =>
  string(5) "en_US"
  'hour' =>
  string(1) "0"
  'minute' =>
  bool(false)
  'second' =>
  bool(false)
}
```
8.0:
```
array(5) {
  'date_format' =>
  string(9) "h:mm:ss a"
  'locale' =>
  string(2) "en"
  'hour' =>
  string(1) "0"
  'minute' =>
  string(0) ""
  'second' =>
  string(0) ""
}
```

As it turns oyt [iconv_substr](https://github.com/php/php-src/pull/6182) behaviour has been normalized and will now always return an empty string if the offset is out of bounds.

We have 73 usages of iconv_substr excluding tests? What's the best way to tackle this? Write a wrapper function that generates the output of iconv_substr <8.0 and replace all calls to iconv_substr with the wrapper function?

falkenhawk:
[...]

PR #32, opened 2020-11-19
PHP8.0 compatibility patch

Megatherium:
Some of the more common things that needed be fixed were:

- A lot of times where PHP used to emit a warning (or even nothing) we
now get Type- and ValueErrors which either need to be prevented by
intializing or casting variables or catching the error.

- file_exists will error out if you feed it NULL bytes. This needed to
be circumvented

- optional parameters need to come after required params, thus some
default values had to be removed from method signatures and workarounds
put in place

- some reflection methods like isArray have become deprecated and are
only called conditionally going forward

- libxml_disable_entity_loader has been deprecated and is only called
conditionally

### DIFF
--- a/packages/zend-barcode/library/Zend/Barcode/Renderer/Image.php
+++ b/packages/zend-barcode/library/Zend/Barcode/Renderer/Image.php
@@ -93,7 +93,7 @@ class Zend_Barcode_Renderer_Image extends Zend_Barcode_Renderer_RendererAbstract
      * Set height of the result image
      *
      * @param null|integer $value
-     * @return Zend_Image_Barcode_Abstract
+     * @return self
      * @throws Zend_Barcode_Renderer_Exception
      */
     public function setHeight($value)
@@ -151,26 +151,25 @@ class Zend_Barcode_Renderer_Image extends Zend_Barcode_Renderer_RendererAbstract
      * Set an image resource to draw the barcode inside
      *
      * @param $image
-     * @return Zend_Barcode_Renderer
+     * @return self
      * @throws Zend_Barcode_Renderer_Exception
      */
     public function setResource($image)
     {
-        if (gettype($image) != 'resource' || get_resource_type($image) != 'gd') {
-            // require_once 'Zend/Barcode/Renderer/Exception.php';
-            throw new Zend_Barcode_Renderer_Exception(
-                'Invalid image resource provided to setResource()'
-            );
+        if ((gettype($image) === 'resource' && get_resource_type($image) === 'gd') || get_class($image) === 'GdImage') {
+            $this->_resource = $image;
+            return $this;
         }
-        $this->_resource = $image;
-        return $this;
+        throw new Zend_Barcode_Renderer_Exception(
+            'Invalid image resource provided to setResource()'
+        );
     }
 
     /**
      * Set the image type to produce (png, jpeg, gif)
      *
      * @param string $value
-     * @return Zend_Barcode_RendererAbstract
+     * @return self
      * @throws Zend_Barcode_Renderer_Exception
      */
     public function setImageType($value)
--- a/packages/zend-file/library/Zend/File/ClassFileLocator.php
+++ b/packages/zend-file/library/Zend/File/ClassFileLocator.php
@@ -58,6 +58,12 @@ class Zend_File_ClassFileLocator extends FilterIterator
 
         parent::__construct($iterator);
         $this->setInfoClass('Zend_File_PhpClassFile');
+
+        if (PHP_VERSION_ID < 80000) {
+            if (!defined('T_NAME_QUALIFIED')) {
+                define('T_NAME_QUALIFIED', 0);
+            }
+        }
     }
 
     /**
@@ -116,6 +122,7 @@ class Zend_File_ClassFileLocator extends FilterIterator
                         switch ($type) {
                             case T_STRING:
                             case T_NS_SEPARATOR:
+                            case T_NAME_QUALIFIED:
                                 $namespace .= $content;
                                 break;
                         }
--- a/packages/zend-filter/library/Zend/Filter/Compress/Bz2.php
+++ b/packages/zend-filter/library/Zend/Filter/Compress/Bz2.php
@@ -152,7 +152,7 @@ class Zend_Filter_Compress_Bz2 extends Zend_Filter_Compress_CompressAbstract
     {
         $archive = $this->getArchive();
         // check $content for NULL bytes or else file_exists will error out
-        if ((0 === preg_match('/\0/', $content)) && @file_exists($content)) {{
+        if ((0 === preg_match('/\0/', $content)) && @file_exists($content)) {
             $archive = $content;
         }
 
--- a/packages/zend-filter/library/Zend/Filter/Compress/Bz2.php
+++ b/packages/zend-filter/library/Zend/Filter/Compress/Bz2.php
@@ -151,7 +151,8 @@ class Zend_Filter_Compress_Bz2 extends Zend_Filter_Compress_CompressAbstract
     public function decompress($content)
     {
         $archive = $this->getArchive();
-        if (@file_exists($content)) {
+        // check $content for NULL bytes or else file_exists will error out
+        if ((0 === preg_match('/\0/', $content)) && @file_exists($content)) {{
             $archive = $content;
         }
 
--- a/packages/zend-filter/library/Zend/Filter/Compress/Gz.php
+++ b/packages/zend-filter/library/Zend/Filter/Compress/Gz.php
@@ -182,7 +182,8 @@ class Zend_Filter_Compress_Gz extends Zend_Filter_Compress_CompressAbstract
     {
         $archive = $this->getArchive();
         $mode    = $this->getMode();
-        if (@file_exists($content)) {
+        // check $content for NULL bytes or else file_exists will error out
+        if ((0 === preg_match('/\0/', $content)) && @file_exists($content)) {
             $archive = $content;
         }
 
--- a/packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
+++ b/packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
@@ -233,7 +233,7 @@ class Zend_Loader_ClassMapAutoloader implements Zend_Loader_SplAutoloader
      * @param  array $parts 
      * @return void
      */
-    public static function resolvePharParentPath($value, $key, &$parts)
+    public static function resolvePharParentPath($value, $key, $parts)
     {
         if ($value !== '...') {
             return;
--- a/packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
+++ b/packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
@@ -208,7 +208,10 @@ class Zend_Loader_ClassMapAutoloader implements Zend_Loader_SplAutoloader
         $prependSlash = $parts && $parts[0] === '' ? '/' : '';
         $parts = array_values(array_filter($parts, array(__CLASS__, 'concatPharParts')));
 
-        array_walk($parts, array(__CLASS__, 'resolvePharParentPath'), $parts);
+        foreach ($parts as $key => $value) {
+            self::resolvePharParentPath($value, $key, $parts);
+        }
+
         if (file_exists($realPath = 'phar://' . $prependSlash . implode('/', $parts))) {
             return $realPath;
         }
@@ -233,7 +236,7 @@ class Zend_Loader_ClassMapAutoloader implements Zend_Loader_SplAutoloader
      * @param  array $parts 
      * @return void
      */
-    public static function resolvePharParentPath($value, $key, $parts)
+    public static function resolvePharParentPath($value, $key, &$parts)
     {
         if ($value !== '...') {
             return;
--- a/packages/zend-locale/library/Zend/Locale/Format.php
+++ b/packages/zend-locale/library/Zend/Locale/Format.php
@@ -929,7 +929,7 @@ class Zend_Locale_Format
                         }
                     } else {
                         // iconv_substr changes since 7.0.11
-                        if (iconv_strlen($splitted[0][0]) === $split) {
+                        if (iconv_strlen($splitted[0][0]) <= $split) {
                             $result['day'] = false;
                         } else {
                             $result['day'] = iconv_substr($splitted[0][0], $split, 2);
@@ -945,7 +945,7 @@ class Zend_Locale_Format
                         }
                     } else {
                         // iconv_substr changes since 7.0.11
-                        if (iconv_strlen($splitted[0][0]) === $split) {
+                        if (iconv_strlen($splitted[0][0]) <= $split) {
                             $result['month'] = false;
                         } else {
                             $result['month'] = iconv_substr($splitted[0][0], $split, 2);
@@ -967,7 +967,7 @@ class Zend_Locale_Format
                         }
                     } else {
                         // iconv_substr changes since 7.0.11
-                        if (iconv_strlen($splitted[0][0]) === $split) {
+                        if (iconv_strlen($splitted[0][0]) <= $split) {
                             $result['year'] = false;
                         } else {
                             $result['year'] = iconv_substr($splitted[0][0], $split, $length);
@@ -984,7 +984,7 @@ class Zend_Locale_Format
                         }
                     } else {
                         // iconv_substr changes since 7.0.11
-                        if (iconv_strlen($splitted[0][0]) === $split) {
+                        if (iconv_strlen($splitted[0][0]) <= $split) {
                             $result['hour'] = false;
                         } else {
                             $result['hour'] = iconv_substr($splitted[0][0], $split, 2);
@@ -1000,7 +1000,7 @@ class Zend_Locale_Format
                         }
                     } else {
                         // iconv_substr changes since 7.0.11 */
-                        if (iconv_strlen($splitted[0][0]) === $split) {
+                        if (iconv_strlen($splitted[0][0]) <= $split) {
                             $result['minute'] = false;
                         } else {
                             $result['minute'] = iconv_substr($splitted[0][0], $split, 2);
@@ -1016,7 +1016,7 @@ class Zend_Locale_Format
                         }
                     } else {
                         // iconv_substr changes since 7.0.11
-                        if (iconv_strlen($splitted[0][0]) === $split) {
+                        if (iconv_strlen($splitted[0][0]) <= $split) {
                             $result['second'] = false;
                         } else {
                             $result['second'] = iconv_substr($splitted[0][0], $split, 2);
--- a/packages/zend-locale/library/Zend/Locale/Math.php
+++ b/packages/zend-locale/library/Zend/Locale/Math.php
@@ -216,11 +216,12 @@ class Zend_Locale_Math
      * Fixes a problem of BCMath with numbers containing exponents
      *
      * @param integer $value Value to erase the exponent
-     * @param integer $scale (Optional) Scale to use
+     * @param int|null $scale (Optional) Scale to use
      * @return string
      */
     public static function exponent($value, $scale = null)
     {
+        $scale = max(0, $scale);
         if (!extension_loaded('bcmath')) {
             return $value;
         }
@@ -242,11 +243,18 @@ class Zend_Locale_Math
      *
      * @param  string  $op1
      * @param  string  $op2
-     * @param  integer $scale
+     * @param  int|null $scale
      * @return string
      */
     public static function Add($op1, $op2, $scale = null)
     {
+        /**
+         * Before PHP8 bcmath functions could be called with a negative
+         * scale factor which was handled as if $scale were 0
+         *
+         * With PHP8 this would emit a ValueError
+         */
+        $scale = max(0, $scale);
         $op1 = self::exponent($op1, $scale);
         $op2 = self::exponent($op2, $scale);
 
@@ -258,11 +266,12 @@ class Zend_Locale_Math
      *
      * @param  string  $op1
      * @param  string  $op2
-     * @param  integer $scale
+     * @param  int|null $scale
      * @return string
      */
     public static function Sub($op1, $op2, $scale = null)
     {
+        $scale = max(0, $scale);
         $op1 = self::exponent($op1, $scale);
         $op2 = self::exponent($op2, $scale);
         return bcsub($op1, $op2, $scale);
@@ -273,11 +282,12 @@ class Zend_Locale_Math
      *
      * @param  string  $op1
      * @param  string  $op2
-     * @param  integer $scale
+     * @param  int|null $scale
      * @return string
      */
     public static function Pow($op1, $op2, $scale = null)
     {
+        $scale = max(0, $scale);
         $op1 = self::exponent($op1, $scale);
         $op2 = self::exponent($op2, $scale);
         return bcpow($op1, $op2, $scale);
@@ -288,11 +298,12 @@ class Zend_Locale_Math
      *
      * @param  string  $op1
      * @param  string  $op2
-     * @param  integer $scale
+     * @param  int|null $scale
      * @return string
      */
     public static function Mul($op1, $op2, $scale = null)
     {
+        $scale = max(0, $scale);
         $op1 = self::exponent($op1, $scale);
         $op2 = self::exponent($op2, $scale);
         return bcmul($op1, $op2, $scale);
@@ -303,11 +314,12 @@ class Zend_Locale_Math
      *
      * @param  string  $op1
      * @param  string  $op2
-     * @param  integer $scale
+     * @param  int|null $scale
      * @return string
      */
     public static function Div($op1, $op2, $scale = null)
     {
+        $scale = max(0, $scale);
         $op1 = self::exponent($op1, $scale);
         $op2 = self::exponent($op2, $scale);
         return bcdiv($op1, $op2, $scale);
@@ -317,11 +329,12 @@ class Zend_Locale_Math
      * BCSqrt - fixes a problem of BCMath and exponential numbers
      *
      * @param  string  $op1
-     * @param  integer $scale
+     * @param  int|null $scale
      * @return string
      */
     public static function Sqrt($op1, $scale = null)
     {
+        $scale = max(0, $scale);
         $op1 = self::exponent($op1, $scale);
         return bcsqrt($op1, $scale);
     }
@@ -345,11 +358,12 @@ class Zend_Locale_Math
      *
      * @param  string  $op1
      * @param  string  $op2
-     * @param  integer $scale
+     * @param  int|null $scale
      * @return string
      */
     public static function Comp($op1, $op2, $scale = null)
     {
+        $scale = max(0, $scale);
         $op1 = self::exponent($op1, $scale);
         $op2 = self::exponent($op2, $scale);
         return bccomp($op1, $op2, $scale);
--- a/packages/zend-pdf/library/Zend/Pdf/Filter/RunLength.php
+++ b/packages/zend-pdf/library/Zend/Pdf/Filter/RunLength.php
@@ -44,30 +44,17 @@ class Zend_Pdf_Filter_RunLength implements Zend_Pdf_Filter_Interface
     {
         $output = '';
 
-        foreach (str_split($data, 128) as $chunk) {
-            $output .= self::chunkEncode($chunk);
-        }
-
-        $output .= "\x80";
-
-        return $output;
-    }
-
-    private static function chunkEncode($data)
-    {
-        $output = '';
-
         $chainStartOffset = 0;
         $offset = 0;
 
         while ($offset < strlen($data)) {
             // Do not encode 2 char chains since they produce 2 char run sequence,
             // but it takes more time to decode such output (because of processing additional run)
-            if (($repeatedCharChainLength = strspn($data, $data[$offset], $offset + 1) + 1)  >  2) {
+            if (($repeatedCharChainLength = strspn($data, $data[$offset], $offset + 1, 127) + 1)  >  2) {
                 if ($chainStartOffset != $offset) {
                     // Drop down previouse (non-repeatable chars) run
                     $output .= chr($offset - $chainStartOffset - 1)
-                        . substr($data, $chainStartOffset, $offset - $chainStartOffset);
+                             . substr($data, $chainStartOffset, $offset - $chainStartOffset);
                 }
 
                 $output .= chr(257 - $repeatedCharChainLength) . $data[$offset];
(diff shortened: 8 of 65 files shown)


glensc:
x-ref:
- https://github.com/zf1s/zf1/issues/31

composer.json
```diff
@@ -16,7 +16,7 @@
         "ext-zlib": "*"
     },
     "require-dev": {
-        "zf1s/phpunit": "^3.7",
+        "zf1s/phpunit": ">=3.7.39",
```
glensc, line 19:
```suggestion
        "zf1s/phpunit": "^3.7.39",
```

lets still keep 3.7 series

packages/zend-barcode/library/Zend/Barcode/Renderer/Image.php
```diff
@@ -151,26 +151,25 @@ public function getWidth()
...
      * @throws Zend_Barcode_Renderer_Exception
      */
     public function setResource($image)
     {
-        if (gettype($image) != 'resource' || get_resource_type($image) != 'gd') {
-            // require_once 'Zend/Barcode/Renderer/Exception.php';
-            throw new Zend_Barcode_Renderer_Exception(
-                'Invalid image resource provided to setResource()'
-            );
+        if ((gettype($image) == 'resource' && get_resource_type($image) == 'gd') || get_class($image) === 'GdImage') {
```
glensc, line 159:
```suggestion
        if ((gettype($image) === 'resource' && get_resource_type($image) === 'gd') || get_class($image) === 'GdImage') {
```

if the line is already changed, upgrade code style to use strict compare

packages/zend-file/library/Zend/File/ClassFileLocator.php
```diff
@@ -68,6 +68,12 @@ public function __construct($dirOrIterator = '.')
                 define('T_NS_SEPARATOR', '\\');
             }
         }
+
+        if (PHP_VERSION_ID < 80000) {
+            if (!defined('T_NAME_QUALIFIED')) {
+                define('T_NAME_QUALIFIED', '');
```
glensc, line 74:
```suggestion
                define('T_NAME_QUALIFIED', 0);
```

those defines are numeric, so let's keep the type?

packages/zend-filter/library/Zend/Filter/Compress/Bz2.php
```diff
@@ -151,7 +151,8 @@ public function compress($content)
     public function decompress($content)
     {
         $archive = $this->getArchive();
-        if (@file_exists($content)) {
+        // check $content for NULL bytes or else file_exists will error out
+        if ((0 === preg_match('/\0/', $content)) && @file_exists($content)) {
```
glensc, line 155:
```suggestion
        if (strpos($content, '/\0/') === false && @file_exists($content)) {
```

use `strpos` and remote superfluous parens

packages/zend-filter/library/Zend/Filter/Compress/Gz.php
```diff
@@ -182,7 +182,8 @@ public function decompress($content)
     {
         $archive = $this->getArchive();
         $mode    = $this->getMode();
-        if (@file_exists($content)) {
+        // check $content for NULL bytes or else file_exists will error out
+        if ((0 === preg_match('/\0/', $content)) && @file_exists($content)) {
```
glensc, line 186:
```suggestion
        if (strpos($content, '/\0/') === false && @file_exists($content)) {
```

packages/zend-locale/library/Zend/Locale/Format.php
```diff
@@ -929,7 +929,7 @@ private static function _parseDate($date, $options)
                         }
                     } else {
                         // iconv_substr changes since 7.0.11
-                        if (iconv_strlen($splitted[0][0]) === $split) {
+                        if (iconv_strlen($splitted[0][0]) <= $split) {
```
glensc, line 932: note that `===` is strict compare but `<=` is not strict compare.

so the semantics will change if both sides are not of same type. have you considered this?

packages/zend-locale/library/Zend/Locale/Math.php
```diff
@@ -216,11 +216,12 @@ public static function localize($value)
      * Fixes a problem of BCMath with numbers containing exponents
      *
      * @param integer $value Value to erase the exponent
-     * @param integer $scale (Optional) Scale to use
+     * @param int|null $scale (Optional) Scale to use
      * @return string
      */
     public static function exponent($value, $scale = null)
     {
+        $scale = $scale > 0 ? $scale : 0;
```
glensc, line 224:
```suggestion
        $scale = max(0, $scale);
```

what you think of using max? and/or perhaps use (int) casting?

packages/zend-reflection/library/Zend/Reflection/Parameter.php
```diff
@@ -58,12 +58,22 @@ public function getDeclaringClass($reflectionClass = 'Zend_Reflection_Class')
      */
     public function getClass($reflectionClass = 'Zend_Reflection_Class')
     {
-        $phpReflection  = parent::getClass();
-        if($phpReflection == null) {
-            return null;
+        if (PHP_VERSION_ID < 80000) {
+            $phpReflection  = parent::getClass();
+            if($phpReflection == null) {
```
glensc, line 63:
```suggestion
            if ($phpReflection == null) {
```

code style

packages/zend-reflection/library/Zend/Reflection/Parameter.php
```diff
@@ -109,13 +119,21 @@ public function getDeclaringFunction($reflectionClass = null)
...
-            if (isset($params[$this->getPosition()])) {
-                return $params[$this->getPosition()]->getType();
-            }
+                if (isset($params[$this->getPosition()])) {
+                    return $params[$this->getPosition()]->getType();
+                }
 
+            }
+        } catch (Zend_Reflection_Exception $e) {
+            if (PHP_VERSION_ID > 79999) {
```
glensc, line 132:
```suggestion
            if (PHP_VERSION_ID >= 80000) {
```

tests/Zend/CodeGenerator/Php/ParameterTest.php
```diff
@@ -134,29 +134,53 @@ public function testFromReflection_GetWithNativeType()
...
-            array('defaultArray', '$array = array ()'),
-            array('defaultArrayWithValues', '$array = array (  0 => 1,  1 => 2,  2 => 3,)'),
-            array('defaultFalse', '$val = false'),
-            array('defaultTrue', '$val = true'),
-            array('defaultZero', '$number = 0'),
-            array('defaultNumber', '$number = 1234'),
-            array('defaultFloat', '$float = 1.34'),
-            array('defaultConstant', '$con = \'foo\'')
-        );
+        if (PHP_VERSION_ID < 80000) {
```
glensc, line 147: perhaps build a temporary array, add conditional things, return the temporary array? this lessens duplication

```php

$result = array(
 // common parts
);

if (PHP_VERSION_ID < 80000) {
$result[]= // specific to old php
}

```

tests/Zend/Date/DateObjectTest.php
```diff
@@ -614,6 +614,20 @@ public function testToStringShouldEqualWithAndWithoutPhpFormat()
             $date->toString(DateTime::ATOM, 'php')
         );
     }
+
+    public function testIncompleteTimeData()
+    {
+        try {
+            $date = new Zend_Date();
+            $date->set(0, Zend_Date::TIMES);
+        } catch (\Exception $e) {
```
glensc, line 623:
```suggestion
        } catch (Exception $e) {
```

I prefer import classes, but as this class is not namespaced, no need to fqdn

tests/Zend/Date/DateObjectTest.php
```diff
@@ -614,6 +614,20 @@ public function testToStringShouldEqualWithAndWithoutPhpFormat()
...
+            $this->fail('This usually stems from not hand iconv_substr propert');
```
glensc, line 624: please use complete English, not geek abbreviations

glensc (changes requested):
Dropped some notes, In overall good work!

tests/Zend/Date/DateObjectTest.php
Megatherium: That's not geek, that's me writing stuff half asleep and not double checking before comitting :)

packages/zend-file/library/Zend/File/ClassFileLocator.php
Megatherium: Doesn't really matter because the sole purpose of the define is to not have undefined constants, but sure why not.

packages/zend-locale/library/Zend/Locale/Format.php
Megatherium: $splitted is `preg_match_all('/\d+/u', $number, $splitted);` so we needn't worry about about its type. $split is either false or int and the false case is always checked explicitly.

packages/zend-locale/library/Zend/Locale/Math.php
Megatherium: Casting would change the behaviour as it would remove any warnings and not return null when using anything else than an int(ish) value

tests/Zend/CodeGenerator/Php/ParameterTest.php
Megatherium: Lessens duplication on the one hand but also makes it slightly less clear as you need keep track of the parts. In the end these tests are still written for humans.

tests/Zend/Date/DateObjectTest.php
glensc: In any case, I'm not able to decipher what that error message means, so it needs to be improved.

packages/zend-locale/library/Zend/Locale/Math.php
```diff
@@ -221,7 +221,7 @@ public static function localize($value)
      */
     public static function exponent($value, $scale = null)
     {
-        $scale = $scale > 0 ? $scale : 0;
+        $scale = max(0, $scale);
```
glensc, line 224: this obviously was for ALL similar changes you made. I would be easier to follow it if you made atomic commits, i.e each of the change as a separate commit. that you could git show and git fixup

highly recommended links to read and follow:
- https://www.freshconsulting.com/atomic-commits/
- https://github.com/keis/git-fixup
- https://github.com/MitMaro/git-interactive-rebase-tool

tests/Zend/CodeGenerator/Php/ParameterTest.php
glensc: But this way I can not make a difference which tests are different for PHP versions unless I go and compare each line by line.

packages/zend-locale/library/Zend/Locale/Math.php
Megatherium: Agreed on the atomic commits. Pure laziness on my side.

Sorry, I have little experience with rebase workflows. Do I understand you correctly that I should rework the fat commit into separate ones via rebasing?

glensc: I would say yes, you should do that, but I'm not able to enforce that on you, and it's definitely not a merge blocker :), besides @falkenhawk has the final say.

I can't give exact help on how to do that, so I've come up with a quick search:
- https://stackoverflow.com/questions/6217156/break-a-previous-commit-into-multiple-commits
- https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_splitting_a_commit

however, I wanted to say here that the comment to use `max` was intended to all similar changes you made `($foo > 0 ? $foo : 0)`, if you don't plan to update all of them, remove the single `max` usage, so the code would be at least consistent.

Megatherium: Totally agreed on the `max`. I'll look into reworking that history. If it looks like a Möbius strip afterwards I'm not taking any responsibility ;)

Megatherium:
Finally got around to reworking that history. Let's see how Travis feels about it :)

Megatherium:
Everything checks out except for that one thing in Zend_Log_Writer_AbstractTest with PHP5.4 which also only occurs if you run the full suite. All three tests pass if you run that thing manually.

Megatherium:
@falkenhawk Do you want me to put a guard in for that specific case so we can have pretty green arrows?

packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
```diff
@@ -237,7 +237,7 @@ public static function concatPharParts($part)
      * @param  array $parts 
      * @return void
      */
-    public static function resolvePharParentPath($value, $key, &$parts)
+    public static function resolvePharParentPath($value, $key, $parts)
```
falkenhawk, line 240: I am not sure here, since the method does not return anything, but modifies $parts inside - I believe the reference must stay here. Does PHP 8 complain here?

packages/zend-pdf/library/Zend/Pdf/Filter/RunLength.php
```diff
@@ -44,17 +44,30 @@ public static function encode($data, $params = null)
     {
         $output = '';
 
+        foreach (str_split($data, 128) as $chunk) {
+            $output .= self::chunkEncode($chunk);
+        }
```
falkenhawk, line 49: could you add a comment here, why the $data has to be split into chunks now, please?

tests/Zend/CodeGenerator/Php/ParameterTest.php
```diff
@@ -134,29 +134,40 @@ public function testFromReflection_GetWithNativeType()
...
+        array('defaultConstant', '$con = \'foo\'')
+    );
+
+        if (PHP_VERSION_ID < 80000) {
+            $commonValues[] = array('hasNativeDocTypes', '$integer');
+        } else {
+            $commonValues[] = array('hasNativeDocTypes', 'int $integer');
+        }
+
+        return $commonValues;
```
falkenhawk, line 170: may I suggest this instead:
```suggestion
        return array(
            array('name', '$param'),
            array('type', 'stdClass $bar'),
            array('reference', '&$baz'),
            array('defaultValue', '$value = \'foo\''),
            array('defaultNull', '$value = null'),
            array('fromArray', 'array $array'),
            array('hasNativeDocTypes', PHP_VERSION_ID >= 80000 ? 'int $integer' : '$integer'),
            array('defaultArray', '$array = array ()'),
            array('defaultArrayWithValues', '$array = array (  0 => 1,  1 => 2,  2 => 3,)'),
            array('defaultFalse', '$val = false'),
            array('defaultTrue', '$val = true'),
            array('defaultZero', '$number = 0'),
            array('defaultNumber', '$number = 1234'),
            array('defaultFloat', '$float = 1.34'),
            array('defaultConstant', '$con = \'foo\'')
        );
```

tests/Zend/Controller/Action/HelperBroker/PriorityStackTest.php
```diff
@@ -57,10 +57,10 @@ public function testStackMaintainsLifo()
...
         $this->stack->push(new Zend_Controller_Action_Helper_Redirector());
         $this->assertEquals(2, count($this->stack));
-        $iterator = $this->stack->getIterator();
-        $this->assertEquals('Zend_Controller_Action_Helper_Redirector', get_class(current($iterator)));
-        next($iterator);
-        $this->assertEquals('Zend_Controller_Action_Helper_ViewRenderer', get_class(current($iterator)));
+        $iterator = $this->stack->getIterator()->getIterator();
+        $this->assertEquals('Zend_Controller_Action_Helper_Redirector', get_class($iterator->current()));
+        $iterator->next();
+        $this->assertEquals('Zend_Controller_Action_Helper_ViewRenderer', get_class($iterator->current()));
```
falkenhawk, line 63: does php8 throw on calling `current()` / `next()` on an iterator?

tests/Zend/Date/DateObjectTest.php
```diff
@@ -614,6 +614,20 @@ public function testToStringShouldEqualWithAndWithoutPhpFormat()
...
+            $date = new Zend_Date();
+            $date->set(0, Zend_Date::TIMES);
+        } catch (Exception $e) {
+            $this->fail('This usually stems from iconv_substr returnvalues in Zend_Locale_Format not being handled properly');
+        } catch (Throwable $e) {
+            $this->fail('This usually stems from iconv_substr returnvalues in Zend_Locale_Format not being handled properly');
+        }
+
+        $this->assertInstanceOf('Zend_Date', $date);
+    }
```
falkenhawk, line 630: new test! yay!

tests/Zend/Db/TestSetup.php
```diff
@@ -80,14 +80,18 @@ protected function _setUpTestUtil()
...
             $conn = $this->_db->getConnection();
         } catch (Zend_Exception $e) {
             $this->_db = null;
             $this->assertTrue($e instanceof Zend_Db_Adapter_Exception,
                 'Expecting Zend_Db_Adapter_Exception, got ' . get_class($e));
             $this->markTestSkipped($e->getMessage());
+        } catch (Throwable $e) {
+            if (stristr($e->getMessage(), 'undefined constant')) {
+                $this->markTestSkipped();
+            }
```
falkenhawk, line 94: In which case this the db factory throws this `undefined constant` error? What kind of constants were missing?

tests/Zend/Filter/HtmlEntitiesTest.php
```diff
@@ -214,7 +214,7 @@ public function testQuoteStyleQuotesEncodeNone()
      */
     public function testCorrectsForEncodingMismatch()
     {
-        if (version_compare(phpversion(), '5.4', '>=')) {
+        if (PHP_VERSION_ID >= 70000 && PHP_VERSION_ID < 70100) {
             $this->markTestIncomplete('Tested feature ZF-11344 is not available because of PHP bug #63450');
         }
```
falkenhawk, line 219: interesting. Was the mentioned bug fixed in later php 5 versions but got reintroduced in php 7.0?

tests/Zend/Form/ElementTest.php
```diff
@@ -1760,12 +1765,12 @@ public function testSetOptionsSetsArrayOfArrayFilters()
...
         $options['filters'] = array(
-            array('Digits', array('bar' => 'baz')),
+            array('Alnum', array('allowWhiteSpace' => true)),
             array('Alpha', array('foo')),
         );
         $this->element->setOptions($options);
-        $filter = $this->element->getFilter('Digits');
-        $this->assertTrue($filter instanceof Zend_Filter_Digits);
+        $filter = $this->element->getFilter('Alnum');
+        $this->assertTrue($filter instanceof Zend_Filter_Alnum);
```
falkenhawk, line 1773: the behavior of this tests has changed. Could you comment on why, please?

tests/Zend/Validate/DateTest.php
```diff
@@ -208,7 +208,11 @@ public function testLocaleContructor()
      */
     public function testNonStringValidation()
     {
-        $this->assertFalse($this->_validator->isValid(array(1 => 1)));
+        try {
+            $this->assertFalse($this->_validator->isValid(array(1 => 1)));
+        } catch (Error $e) {
+            $this->assertTrue($e instanceof TypeError);
+        }
```
falkenhawk, line 215: if validators throw `TypeError`s in PHP8 instead of returning boolean value, I suppose it's a different behavior now? Could it be adjusted in the validator's base class so that it does not throw? Or was the error behavior also expected in php < 8?

tests/Zend/Log/Writer/StreamTest.php
```diff
@@ -55,6 +55,8 @@ public function testConstructorThrowsWhenResourceIsNotStream()
         } catch (Exception $e) {
             $this->assertTrue($e instanceof Zend_Log_Exception);
             $this->assertRegExp('/not a stream/i', $e->getMessage());
+        } catch (TypeError $e) {
+            $this->assertRegExp('/must be of t/i', $e->getMessage());
```
falkenhawk, line 59: I am not sure about the adjustments here. Were these `TypeError`, `ValueError` triggered as regular errors in PHP <8 or were they simply non-errors before and now PHP 8 is more strict? If that's the case, is it okay that PHP8 will throw such errors instead of the php <8 shrug-and-proceed behavior?

falkenhawk:
> 
> 
> @falkenhawk Do you want me to put a guard in for that specific case so we can have pretty green arrows?

@Megatherium I will try to run php5.4 test suite locally to see if I can figure out why it fails

tests/Zend/Db/TestSetup.php
Megatherium: DB2 constants

tests/Zend/Filter/HtmlEntitiesTest.php
Megatherium: I don't really know. I was probably playing around and saw that the bug only occurred on 7.0

tests/Zend/Log/Writer/StreamTest.php
Megatherium: Usually where I added Error catches the behaviour in <8 was to emit a warning or fail silently.

tests/Zend/Controller/Action/HelperBroker/PriorityStackTest.php
Megatherium: It retuns false and thus get_class throws a TypeError

packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
Megatherium: PHP8 does complain that $parts is value when the function is called from `realPharPath`. I assumed PHP8 had changed something fundamentally about pass-by-ref. Turns out that it's only `array_walk` that's changed and it now automatically passes the $userdata param by value meaning that this can be circumvented by changing array_walk to a foreach loop.

tests/Zend/Form/ElementTest.php
Megatherium: The behaviour hasn't really changed because the intent of the test is to use an array of arrays in `setOptions`. What has changed is that PHP8 supports calling functions with named parameters and will throw a Fatal if you try to access a parameter by name that does not exist. `Digits::__construct` has no param named bar ergo we get a Fatal.

I could add a version switch but that wouldn't make the test any more meaningful.

tests/Zend/Validate/DateTest.php
Megatherium: The TypeError isn't thrown by any validator. It's due to using an array as the first argument in `iconv_strpos`. Before 8 this just emitted a warning and returned a false.

tests/Zend/CodeGenerator/Php/ParameterTest.php
Megatherium: Fine by me

packages/zend-pdf/library/Zend/Pdf/Filter/RunLength.php
Megatherium: To work around this [bug](https://github.com/php/php-src/commit/d776d25a8e6763bfe8cc283a4693d7d417d50ddd). Which isn't in the 8.0.0 release.

falkenhawk: meanwhile travis is still fighting with setting up the 8.0 image https://travis-ci.community/t/php-8-0-missing/10132/14

tests/Zend/Validate/DateTest.php
falkenhawk: but `iconv_strpos` is called inside the date validator, isn't it? Perhaps we could check the type of the value to be validated inside the validator and return false if type is wrong, to avoid the `TypeError` to be thrown and have the behavior consistent among all php versions? 🤔

tests/Zend/Log/Writer/StreamTest.php
falkenhawk: Okay I guess it is fine in case of this component - when something is wrong with internals of this stream thing, it should throw

tests/Zend/Validate/DateTest.php
Megatherium: OK. In `Zend_Validate_Date::_checkFormat` exceptions are already caught and the method returns false. This just needed to be extended to also catch errors. Test has been reverted.

glensc:
https://github.com/zf1s/zf1/issues/35 created a separate issue, so it can be worked independently of this PR.

Megatherium:
So, next step:

Rebase this to master, add 8.0 to Github actions, disallow failure for 8.0 on Travis.

Correct?

Megatherium:
I don't really get what that test suddenly bails in 5.4. As far as I can tell I haven't touched any file involved in that test. Also that test is supposed to cause an error which PHPUnit should handle... I'm confused.

Megatherium:
I'm tempted to simply delete the offending lines because in the end they don't even really test the code but rather if PHP's type system works.

Opinions? @glensc @falkenhawk

glensc:
> I'm tempted to simply delete the offending lines because in the end they don't even really test the code but rather if PHP's type system works.

I'm not on the page with problems in this PR, it's too big to even find what is the question.

Aside, the other fork took changes from here and merged and made release:
- https://github.com/Shardj/zf1-future/pull/130

falkenhawk:
So PHP5.4, and only 5.4, on GHA throws:
```
PHP Catchable fatal error:  Argument 1 passed to Zend_Log_Writer_Abstract::setFormatter() must implement interface Zend_Log_Formatter_Interface, instance of stdClass given, called in /home/runner/work/zf1/zf1/tests/Zend/Log/Writer/AbstractTest.php on line 68 and defined in /home/runner/work/zf1/zf1/packages/zend-log/library/Zend/Log/Writer/Abstract.php on line 97
```
the offending line is https://github.com/zf1s/zf1/blob/674324a33d4f7097654d81f9fcffa332cf5b7fa3/tests/Zend/Log/Writer/AbstractTest.php#L68
and then also identical at line 77.

The question here is why is only 5.4@GHA throwing. Could it be an issue specific to that php build that it's more strict than other 5.x versions? (actually behaving like php7+) 🤔

Megatherium:
It's not only GHA. [Travis](https://travis-ci.org/github/Megatherium/zf1/jobs/758030508) also shits the bed and if I run the whole suite locally in Docker (minity/php54-console) but ONLY if I run the whole suite (tests/Zend/Alltests.php). If I run Log/AllTests.php, Log/Writer/AllTests.php or Log/Writer/AbstractTest.php directly all tests pass.

composer.json
```diff
@@ -19,7 +19,7 @@
         "symfony/polyfill-php70": "^1.19"
     },
     "require-dev": {
-        "php-parallel-lint/php-parallel-lint": "1.0.0",
+        "php-parallel-lint/php-parallel-lint": "^1.0.0",
```
glensc, line 22: Remove this change. If you want this change, submit in separate PR for discussion.

falkenhawk:
@Megatherium I'd personally remove / skip these two tests - it does not add any value for testing a language-specific feature (type-checking) imo. I can't guess what makes the 5.4 shudder and barf on that but only when running all tests - I could keep wondering what from other tests could mutate the behavior globally... but I don't have time for this ;)

composer.json
glensc: Separated from this PR:
- https://github.com/zf1s/zf1/pull/58

glensc:
I'm having a look at the Zend_Log_Writer_AbstractTest but first is this trouble:
- https://github.com/zf1s/zf1/pull/58

composer.json
Megatherium: Removed

Megatherium:
Something's fishy:

https://github.com/Megatherium/zf1/blob/fd0c379b1217388fb848091ed342e974ee86c2af/packages/zend-filter/library/Zend/Filter/HtmlEntities.php#L207

returns false under 5.3 in the HtmEntitiesTest...

glensc:
> Something's fishy:
> 
> [Megatherium/zf1@`fd0c379`/packages/zend-filter/library/Zend/Filter/HtmlEntities.php#L207](https://github.com/Megatherium/zf1/blob/fd0c379b1217388fb848091ed342e974ee86c2af/packages/zend-filter/library/Zend/Filter/HtmlEntities.php#L207)
> 
> returns false under 5.3 in the HtmEntitiesTest...

for what `$enc` value?

Megatherium:
The contents of tests/Zend/Filter/_files/latin-1-text.txt:

`Here is a dash: "<97>"` (copied from vim)

Which looks borked , but isn't the whole intent of `//IGNORE` to basically just get a 'best effort'?

glensc:
I mean `$enc` value:
- https://github.com/Megatherium/zf1/blob/fd0c379b1217388fb848091ed342e974ee86c2af/packages/zend-filter/library/Zend/Filter/HtmlEntities.php#L206

Megatherium:
> I mean `$enc` value:
> 
>     * https://github.com/Megatherium/zf1/blob/fd0c379b1217388fb848091ed342e974ee86c2af/packages/zend-filter/library/Zend/Filter/HtmlEntities.php#L206

Oh, sorry. UTF-8

glensc:
Created test:
- https://3v4l.org/MU8Yr

The errors go let to right, very confusing

- If I launch live preview it shows with 7.4.0: `Notice: iconv(): Unknown error (84) in /tmp/preview on line 17`
- If I do live preview buy manually selecting 7.4.0, it's ok

also, seems php 5.4 is built without iconv there.

glensc:
@falkenhawk @Megatherium seems this is stalled again. Is there any interest, I could cherry-pick commits related from here to separate PRs? So the changeset could be merged individually, not depending on the "big approval".

I need confirmation, so I won't end up doing unwanted work.

falkenhawk:
@glensc thanks for stepping up. I believe there is lots of interest, but I guess we all might be constrained by the regular job responsibilities or other stuff. As much as I would like to have it done, I can't merge it yet when there are tests failing. I would also love to jump in and help but I am afraid I can't do at the moment.

Is that failing test due to the issue with iconv the only thing left to be sorted out here? If you have an idea how to fix it then perhaps could you either rebase or merge from master and push a fix? 🙏

Please coordinate with @Megatherium if possible so that work is not doubled, perhaps Alex has some work not pushed yet.

Megatherium:
The thing is I believe the test setup and not the code to be wrong. I know this is basically like claiming a compiler bug but iconv is an external resource and I didn't have my little grubby hands anywhere near the failing code.
Same with Zend_Queue_Queue1Test: if you run it or the whole Zend_Queue test battery you don't get any failures. Only if you run tests/Zend/AllTests.php do you see that error and that means (at least to me, although I might be lacking some insight here) that we have some kind weird side effects that have nothing to do with the framework code.

tl;dr: I currently don't know how to give all green check marks.

glensc:
@falkenhawk @Megatherium I proposed to extract all changes that are ok to a separate PR. I would not be interested in fixing actual bugs right now. I would like to receive the clear Go Ahead signal. Or would you @Megatherium want to do that yourself?

falkenhawk:
@glensc please do 🙏

Megatherium:
@glensc Please do. Maybe by chopping this stuff up it'll help me to narrow down the source of the remaining problems.

glensc:
Created 14 PR's as can see from the web:
- https://github.com/zf1s/zf1/pull/32#issuecomment-782047133

The last one seems to have issues, probably depends on the change that went to some other pr:
- https://github.com/zf1s/zf1/pull/77

or it could be just a flaky test.

falkenhawk:
Closing in favor of multitude PRs from @glensc . Thank you both!

PR #34, opened 2020-12-03
Replace Travis with GitHub Actions

glensc:
- [x] GHA Jobs not appearing under Checks. Probably needs to be merged to `master` first.
- [x] [Zend_Ldap_ConverterTest::testToLdapSerialize](https://github.com/zf1s/zf1/pull/34#issuecomment-737771053): Fixed by using `CI` instead of `TRAVIS` env
- [ ] [Zend_Db_Table_Select_MysqliTest::testSelectWhereWithTypeFloat
Zend_Db_Select::_join(): Error occurred while closing statement](https://github.com/zf1s/zf1/pull/34#issuecomment-737886457)
- [ ] [Zend_Locale_FormatTest::testToFloatSetlocale](https://github.com/zf1s/zf1/pull/34#issuecomment-739896070)

PR #37, opened 2020-12-07
Setup GitHub Actions to lint all files in the repository

glensc:
Setup GitHub Actions to lint all files in the repository to catch issues like https://github.com/zf1s/zf1/pull/36

The output is very noisy, so therefore it's set up as a separate job.

Extracted base GitHub Action setup from https://github.com/zf1s/zf1/pull/34

In the future could use some other tools like:
- https://github.com/overtrue/phplint
- https://github.com/JakubOnderka/PHP-Parallel-Lint

but let's keep this simple, let's have this as a first GitHub action enabled in this repo.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,8 @@ jobs:
       - name: Lint PHP source files
         run: |
           # Lint All PHP source files
+          # Remove source expected to have syntax error
+          git rm tests/Zend/Loader/_files/ParseError.php
           git ls-files '**/*.php' | xargs -l1 php -l
 
 # vim:ft=yaml:et:ts=2:sw=2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,4 +32,9 @@ jobs:
         with:
           php-version: ${{ matrix.php }}
 
+      - name: Lint PHP source files
+        run: |
+          # Lint All PHP source files
+          git ls-files '**/*.php' | xargs -l1 php -l
+
 # vim:ft=yaml:et:ts=2:sw=2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,6 +37,7 @@ jobs:
           # Lint All PHP source files
           # Remove source expected to have syntax error
           git rm tests/Zend/Loader/_files/ParseError.php
-          git ls-files '**/*.php' | xargs -l1 php -l
+          git ls-files '**/*.php' | while read f; do php -l "$f" > /dev/null || touch failed && echo -n .; done
+          test ! -f failed
 
 # vim:ft=yaml:et:ts=2:sw=2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,35 @@
+name: "Tests"
+
+on:
+  - pull_request
+  - push
+
+jobs:
+  test:
+    name: PHP ${{ matrix.php }}
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        php:
+          - "5.3"
+          - "5.4"
+          - "5.5"
+          - "5.6"
+          - "7.0"
+          - "7.1"
+          - "7.2"
+          - "7.3"
+          - "7.4"
+          - "8.0"
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Setup PHP ${{ matrix.php }}
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+
+# vim:ft=yaml:et:ts=2:sw=2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,6 +31,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
+          ini-values: display_errors=off, log_errors=on
 
       - name: Lint PHP source files
         run: |


glensc:
requires https://github.com/zf1s/zf1/pull/36 to be merged first (will rebase once it's merged)

falkenhawk:
Love it! It caught another issue on PHP 8.0:
```
PHP Fatal error:  Unparenthesized `a ? b : c ? d : e` is not supported. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)` in packages/zend-service-console/library/Zend/Service/Console/Command.php on line 223
Errors parsing packages/zend-service-console/library/Zend/Service/Console/Command.php
```

Would it be possible to silence all the success `No syntax errors detected in ...` messages? 🤔

glensc:
I guess the future is now then:

> In the future could use some other tools like:
> - https://github.com/overtrue/phplint
> - https://github.com/JakubOnderka/PHP-Parallel-Lint

I've personally used https://github.com/JakubOnderka/PHP-Parallel-Lint in the past, so I'll switch using that.

falkenhawk:
> Would it be possible to silence all the success `No syntax errors detected in ...` messages? 🤔

https://stackoverflow.com/a/34214376/3729316

glensc:
Needed to adjust ini settings for errors to be in stderr, messages in stdout.

Something I researched a decade ago:
- https://docs.google.com/document/d/e/2PACX-1vQXNVeuGGu8WHJLgqyf4RxpaByIqKh_CXKLUq1CKCreW2W70NRE_1h7d9IICEiKFsUWQ5LNHldbkp1f/pub

seems PHP defaults are still silly.

glensc:
Doh, adding shell pipe lost the ability to report exit code:
- https://github.com/glensc/php-zf1s/runs/1516677207?check_suite_focus=true

the output looks compact now, command failed, but exit code is `0`:

```
Run # Remove source expected to have syntax error
rm 'tests/Zend/Loader/_files/ParseError.php'
PHP Fatal error:  Unparenthesized `a ? b : c ? d : e` is not supported. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)` in packages/zend-service-console/library/Zend/Service/Console/Command.php on line 223
xargs: php: exited with status 255; aborting
............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
```

glensc:
this will be ready after https://github.com/zf1s/zf1/pull/39 merge

glensc:
Removed draft, ready for merge.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+
+      - name: Lint PHP source files
+        run: |
+          # Lint All PHP source files
+          # Remove source expected to have syntax error
+          git rm tests/Zend/Loader/_files/ParseError.php
+          git ls-files '**/*.php' | (xargs -l1 php -d display_errors=off -d log_errors=on -l && touch ok) | while read l; do echo -n .; done
```
falkenhawk, line 40: are these ini settings really necessary? I could see the error in PHP 8.0 job was already outputted prior to the change from https://github.com/glensc/php-zf1s/commit/895f2d3e49c500447b2f3d53c1c87103f8d11296 - https://github.com/glensc/php-zf1s/runs/1516661591
```suggestion
          git ls-files '**/*.php' | (xargs -l1 php -l && touch ok) | while read l; do echo -n .; done
```

glensc: yes, because without them the errors are printed to stdout as well, and stdout is used for counting files (printing dots).

falkenhawk: btw do you maybe know of any way to make it continue even on errors encountered, so that it would check all files and print all errors, instead of printing just the first one and bail? 🤔

falkenhawk (approved):

.github/workflows/tests.yml
glensc: it should test all files, I see no reason how it could fail, `xargs` invokes `php -l` for each file (`xargs -l`).

Altho I have not explicitly tested that is true because of the above reasons.

falkenhawk: I checked https://github.com/glensc/php-zf1s/runs/1516737279?check_suite_focus=true and compared it with 7.4 job and noticed that there are less dots in the 8.0 job - meaning it most probably stops on first error

glensc: Indeed, looks like `xargs` behavior is to abort:

```
xargs: php: exited with status 255; aborting
```

I could not find an option from `xargs` manual page to change this behavior, so I'll use shell `while` loop instead.

falkenhawk: I also found this https://github.com/StephaneBour/actions-php-lint/blob/7.4/entrypoint.sh - maybe this would be of any help.
Apparently there is an action ready to be used, but the author went for linking its version to php version used in `Dockerfile` and 8.0 is not available yet.

glensc: OK, looks better now:
- https://github.com/glensc/php-zf1s/runs/1523182249?check_suite_focus=true

glensc: Alternatives to using not our own code are listed in PR body and also in a note:
- https://github.com/zf1s/zf1/pull/37#issuecomment-740522629

The actions I've looked, had a specific one php version only, invoked docker, but we already have php in place. so these should be ran from composer, but I've planned composer setup in next PR as there are issues with php 8 to solve.

glensc: There's some optimization point too, add `-n` to skip loading ini files:
- https://github.com/phingofficial/phing/commit/8cfdd3ae293cb14c0e00816d6047d717153f6c9e

but it has a downside that error reporting needs to be reset or funny things happen with short open tag files. it's addressed in later commits in phing, but their trac being decommissioned it's harder to find those details.

glensc: Anyway, this is the latest one from this branch with two files with errors:
- https://github.com/glensc/php-zf1s/runs/1523341477?check_suite_focus=true

It's ready for merge from my side.

glensc: And job matching this branch:
- https://github.com/glensc/php-zf1s/runs/1523337914?check_suite_focus=true

glensc:
yay, 🚢 :shipit:

glensc:
What I wanted to see, is that the actions now appear in zf1s/zf1 actions tab:
- https://github.com/zf1s/zf1/actions/

And it does:
- https://github.com/zf1s/zf1/actions/runs/410431517

What I don't like at all is that you squashed my carefully crafted commits to a single commit!

PR #38, opened 2020-12-08
Convert all *.php files to Unix line encoding

glensc:
Required to solve syntax error in one of these files:
- https://github.com/zf1s/zf1/pull/37#issuecomment-740449838

```
git ls-files '**/*.php' | xargs dos2unix
```


glensc:
Had real trouble fixing the syntax error as could not find an editor that would not try to convert file encoding when making a single line change.

In vim 'set bin' finally helped, but I think the file encodings should be unified really.

glensc:
cc @falkenhawk 

not sure should I ping you in MR's or you have all notifications enabled in this repository.

falkenhawk:
I have all notifications enabled.
Thank you for yet another PR!

falkenhawk (approved):

glensc:
@falkenhawk ok great, I'll not bother pinging you separately then. If PR is not in draft state, it's ready to be merged if it's ok for you too.

PR #41, opened 2020-12-22
GitHub Actions: Install composer dependencies

glensc:
_Extracted composer install from https://github.com/zf1s/zf1/pull/34_

- Use versioned dependencies in composer.json
- Install composer dependencies in CI
- Fix validation of main composer.json
- Validate also main composer.json

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,7 +45,7 @@ jobs:
       # https://github.com/zf1s/zf1/pull/6#issuecomment-495397170
       - name: Validate composer.json for all packages
         run: |
-          for json in packages/*/composer.json; do COMPOSER=$json composer validate || touch failed; done
+          for json in composer.json packages/*/composer.json; do COMPOSER=$json composer validate || touch failed; done
           test ! -f failed
 
 # vim:ft=yaml:et:ts=2:sw=2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,7 +32,7 @@ jobs:
         with:
           php-version: ${{ matrix.php }}
           ini-values: display_errors=off, log_errors=on
-          extensions: :xdebug
+          extensions: :xdebug, ctype, dom, gd, iconv, ldap, mbstring, reflection, session, simplexml, spl, xml, zlib, memcache, pdo_sqlite, bcmath
 
       - name: Lint PHP source files
         run: |
@@ -47,5 +47,8 @@ jobs:
         run: |
           for json in composer.json packages/*/composer.json; do COMPOSER=$json composer validate || touch failed; done
           test ! -f failed
+
+      - name: Install dependencies
+        run: composer install --prefer-dist --no-progress --no-suggest
 
 # vim:ft=yaml:et:ts=2:sw=2
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,7 @@
 {
     "name": "zf1s/zf1",
+    "description": "Monorepo for zf1s (Zend Framework 1) packages",
+    "license": "BSD-3-Clause",
     "require": {
         "php": ">=5.3.3",
         "ext-ctype": "*",
--- a/composer.json
+++ b/composer.json
@@ -18,8 +18,8 @@
         "ext-zlib": "*"
     },
     "require-dev": {
-        "zf1s/phpunit": "^3.7",
-        "phpunit/dbunit": "1.3.*"
+        "zf1s/phpunit": "3.7.39",
+        "phpunit/dbunit": "1.3.2"
     },
     "autoload": {
         "psr-0": {


.github/workflows/tests.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         with:
           php-version: ${{ matrix.php }}
           ini-values: display_errors=off, log_errors=on
-          extensions: :xdebug
+          extensions: :xdebug, ctype, dom, gd, iconv, ldap, mbstring, reflection, session, simplexml, spl, xml, zlib, memcache, pdo_sqlite, bcmath
```
glensc, line 35: somewhy ext-ldap is missing for 5.3 (you need to click on "Setup PHP 5.3" to see output):
- https://github.com/glensc/php-zf1s/runs/1597658212?check_suite_focus=true

```
==> Setup Extensions
✓ ctype Enabled
✓ dom Enabled
✓ gd Enabled
✓ iconv Enabled
✗ ldap Could not install ldap on PHP 5.3.29
✓ mbstring Enabled
...
```

glensc: btw, some earlier builds have succeeding as the cache is used and composer install step skipped:
- https://github.com/glensc/php-zf1s/runs/1597629716?check_suite_focus=true

.github/workflows/tests.yml
```diff
@@ -45,7 +45,20 @@ jobs:
...
-          for json in packages/*/composer.json; do COMPOSER=$json composer validate || touch failed; done
+          for json in composer.json packages/*/composer.json; do COMPOSER=$json composer validate || touch failed; done
           test ! -f failed
 
+      - name: Cache Composer packages
+        id: composer-cache
+        uses: actions/cache@v2
+        with:
+          path: vendor
+          key: ${{ runner.os }}-php-${{ hashFiles('**/composer.lock') }}
```
glensc, line 56: as the composer.lock is not tracked in git, not sure if should use composer.json here instead?

as the docs don't mention that:
- https://github.com/actions/cache/blob/9f3a4d3e65eb0e37d28bc055c0d1a91eac6d76fe/examples.md#php---composer

but I've seen some examples saying to use package.json.

probably running this action several times gives the answer....

glensc: here's some discussion on the not committed lock file:
- https://github.com/bamarni/composer-bin-plugin/issues/50#issuecomment-604512815

falkenhawk: I think pointing to `composer.json` would make sense if versions specified in composer.json were fixed. (no version ranges)
E.g. in case new version of `zf1s/phpunit` is released, `actions/cache` may still take previous version out of its cache based on the hash of `composer.json`.

Does it make sense to generate and commit lockfile here? Then, when a new version of `zf1s/phpunit` is released, another commit will have to be pushed to `zf1s/zf1` to update the version in the lockfile.
Is there any other option? 🤔

glensc: My original point was that if composer.json has changes, and it's hash not used for actions/cache, the vendor would not be updated.

If you use composer.json without locking, it's expected to have a cache with a variable contents.

I'm not sure about committing a lock file here. I'll give it a thought.

.github/workflows/tests.yml
glensc: reported: https://github.com/shivammathur/setup-php/issues/386

glensc: https://github.com/shivammathur/setup-php/issues/386 is fixed

.github/workflows/tests.yml
glensc: ## Lock file for composer.json

There's a common misconception whether `composer.lock` needs to be committed to the repository or not.

It can be summarized with two opposite scenarios:

- projects: [keep lock] in projects where it really matters (production is deployed from it):
- libraries: [do not keep lock] for libraries

[keep lock]: https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control
[do not keep lock]: https://getcomposer.org/doc/02-libraries.md#lock-file

As the current composer.json contains only one dependency, let's make it strict (no version ranges) and do not commit lock. as the lock would be just tracking phpunit dependencies, that's silly in my opinion.

And we can rethink this in the future at any time.

As for any other option: don't use `actions/cache` and such problem won't exist :)

.github/workflows/tests.yml
```diff
@@ -45,7 +45,24 @@ jobs:
...
+      - name: Cache Composer packages
+        id: composer-cache
+        uses: actions/cache@v2
+        with:
+          path: vendor
+          # hashFiles is path relative to GITHUB_WORKSPACE
+          # we hash only root composer.json
+          # https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#hashfiles
+          # https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
+          key: ${{ runner.os }}-php-${{ hashFiles('/composer.json') }}
```
glensc, line 60: now I'm thinking perhaps the different php versions should not share composer cache:
- https://github.com/zf1s/zf1/actions/runs/461660032

otherwise, they end up not being ran at all, as cache created from other run, which is possibly different PHP version... which could end up shadowing some bugs.

but then again vendor should be identical in regardless what was the PHP version when composer install was invoked. there's one gotcha: composer v2 generates different autoload if [platform-check](https://getcomposer.org/doc/06-config.md#platform-check) is enabled (and it is by default).


@falkenhawk wdyt?

glensc: made the change to be able to proceed with merging. can easily remove the commit if it's unwanted.

.github/workflows/tests.yml
falkenhawk: I think it's a good idea to lock dependencies to exact versions in composer.json in this case.

falkenhawk: Btw how much time is saved in a run when composer cache is hit?

glensc: please look at the changes, it's already implemented

falkenhawk: I saw them, only confirmed that's a good idea :) Thanks.

falkenhawk: > 
> 
> Btw how much time is saved in a run when composer cache is hit?

To answer my question: I checked https://github.com/glensc/php-zf1s/actions/runs/466000519 and I could see composer install takes 5-10 seconds, when cache is not hit, so in our (simple) case with only 2 dependencies maybe it's not worth to add additional complexity here 🤔

Also something is wrong with `hashFiles`: the hash is missing from the cache key
image

glensc: Removed all cache related changes

glensc: Altho I personally would stick using the cache, then CI would not have to make requests outside its infrastructure. But then again I would like this merged, so all cache logic is removed now.

falkenhawk (approved):
Thank you once again for your contributions.

PR #44, opened 2021-01-11
Add minimal phpunit run in GitHub Actions

glensc:
Adds minimal PHPUnit invocation. Most of the tests are Skipped.

Tests PHP 5.3-7.4, PHP 8.0 is removed.

Extracted from https://github.com/zf1s/zf1/pull/34

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,4 +50,7 @@ jobs:
       - name: Install dependencies
         run: composer install --prefer-dist --no-progress --no-suggest
 
+      - name: Run PHPUnit tests
+        run: vendor/bin/phpunit --verbose
+
 # vim:ft=yaml:et:ts=2:sw=2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,7 +21,6 @@ jobs:
           - "7.2"
           - "7.3"
           - "7.4"
-          - "8.0"
 
     steps:
       - name: Checkout Code
--- a/tests/Zend/Cache/CommonBackendTest.php
+++ b/tests/Zend/Cache/CommonBackendTest.php
@@ -232,7 +232,7 @@ abstract class Zend_Cache_CommonBackendTest extends PHPUnit_Framework_TestCase {
     public function testCleanModeAll()
     {
         if ($this instanceof Zend_Cache_MemcachedBackendTest
-            && getenv('TRAVIS')
+            && getenv('CI')
         ) {
             $this->markTestSkipped(
                 'Test randomly fail on Travis CI.'
--- a/tests/Zend/Cache/FileBackendTest.php
+++ b/tests/Zend/Cache/FileBackendTest.php
@@ -161,7 +161,7 @@ class Zend_Cache_FileBackendTest extends Zend_Cache_CommonExtendedBackendTest {
 
     public function testSaveWithABadCacheDir()
     {
-        if (getenv('TRAVIS')) {
+        if (getenv('CI')) {
             $this->markTestSkipped(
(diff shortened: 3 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -12,15 +12,6 @@ jobs:
...
-          - "5.3"
-          - "5.4"
-          - "5.5"
-          - "5.6"
-          - "7.0"
-          - "7.1"
-          - "7.2"
-          - "7.3"
-          - "7.4"
           - "8.0"
```
glensc, line 15: Something funny happens, PHPUnit shows progress until 89%:
- https://github.com/glensc/php-zf1s/runs/1685017909?check_suite_focus=true

but the build is success! this is not good.

glensc: I've ended up running same tests locally via docker, and there phpunit stops at 90%

glensc: @falkenhawk what you think of disabling 8.0 for github actions, I ended up cherry-picking commits from the 8.0 branch, and seems it's still having issues. I don't want to make the PR itself very big, but rather make several PR's that are easier review and merge. possibly doing things in parallel, once there's basic phpunit invocation present.

or alternatively, can just ignore errors from 8.0 phpunit, merge with failing pipeline.

I propose making this PR exclude any issues with 8.0 (deal them later), either by removing 8.0 it consciously ignoring errors from 8.0 job.

ps: github actions does not have "allow_failure" equivalent yet. I've been tracking this issue for it: https://github.com/actions/toolkit/issues/399.

your input needed how to proceed.

glensc: my next PR plans(after this PR) are (from https://github.com/zf1s/zf1/pull/34):
- add locale tests and locale fixes
- add services, mysql, pgsql

falkenhawk: wow I greatly appreciate your dedication.
Let's skip 8.0 for now and ditch any compatibility fixes needed for 8.0 here and deal with that in a separate PR, possibly branched off the 8.0 compatibility patch branch.

Having the whole pipeline running for php 5.3-7.4 on GHA would be awesome, as the first step. I asked travis for free OSS credits 2 weeks ago - because the jobs are currently not running at all due to depleted credits - and still no answer 🤷‍♂️

glensc: Oh, didn't know Travis OSS credits have been limited, I'm not following the news apparently.

glensc:
Updated to test only 5.3 to 7.4

falkenhawk (approved):
Thank you!

btw are `stty: 'standard input': Inappropriate ioctl for device` printed out during tests because of `--verbose` phpunit flag, or is it something else, yet to be figured out?

glensc:
haven't checked those `stty` messages. Since these do not trigger errors, I assumed they've always been there.

falkenhawk:
I haven't seen them before, but it's okay to check them later.

.github/workflows/tests.yml
glensc: Seems that enabling locales broken sql tests:
- https://github.com/zf1s/zf1/pull/34#issuecomment-739904178

so, need to enable sql testing next, then locales testing

glensc:
@falkenhawk please could you not do squash merges, please!

falkenhawk:
Okay I will keep that in mind, but could you tell me please why you don't like it? In this case I thought it's fine to have only one nice commit

glensc:
@falkenhawk, I follow:
- https://www.freshconsulting.com/atomic-commits/
- https://chris.beams.io/posts/git-commit/

PR #45, opened 2021-01-12
Restore locales before calling test assertions

glensc:
Restore locales before calling assertions otherwise sql will be broken:

From https://github.com/zf1s/zf1/pull/34#issuecomment-739904178:
> seems enabling locale tests by installing needed locales, brings more trouble than needed.
> 
> the tests are not run in proper isolation, i.e global resource like locale must be restored after a test, and that is not performed, neither are tests run in isolation (i think phpunit 3.7 does not even support that), and sql queries are written with very bad style, resulting syntax error as numeric value value formatted with comma according to locale rules, gets parsed as two values in sql.
> 
> ```
> 2) Zend_Db_Table_Select_MysqliTest::testSelectWhereOr
> Zend_Db_Exception: SQL error for "INSERT INTO `zfprice` (`product_id`, `price_name`, `price_total`) VALUES (1, 'Price 1', 200,45)": Column count doesn't match value count at row 1
> ```
> 
> - https://github.com/glensc/php-zf1s/runs/1510708630?check_suite_focus=true
> 

NOTE: fixing the DB layer to be independent on locale is a legacy that zf1 carries and can't be fixed.

### DIFF
--- a/tests/Zend/Locale/FormatTest.php
+++ b/tests/Zend/Locale/FormatTest.php
@@ -36,6 +36,12 @@
 class Zend_Locale_FormatTest extends PHPUnit_Framework_TestCase
 {
     /**
+     * Constant for Non-breaking space UTF-8 encoded value.
+     * https://en.wikipedia.org/wiki/Non-breaking_space
+     */
+    const NBSP = " ";
+
+    /**
      * teardown / cleanup
      */
     public function tearDown()
@@ -944,7 +950,7 @@ class Zend_Locale_FormatTest extends PHPUnit_Framework_TestCase
         $myFloat = 1234.5;
         $test1 = Zend_Locale_Format::toFloat($myFloat, $params_fr);
         $test2 = Zend_Locale_Format::toFloat($myFloat, $params_en);
-        $this->assertEquals("1 234,50", $test1);
+        $this->assertEquals("1" . self::NBSP . "234,50", $test1);
         $this->assertEquals("1,234.50", $test2);
         // placing tearDown here (i.e. restoring locale) won't work, if test already failed/aborted above.
     }
--- a/tests/Zend/Locale/FormatTest.php
+++ b/tests/Zend/Locale/FormatTest.php
@@ -39,7 +39,7 @@ class Zend_Locale_FormatTest extends PHPUnit_Framework_TestCase
      * Constant for Non-breaking space UTF-8 encoded value.
      * https://en.wikipedia.org/wiki/Non-breaking_space
      */
-    const NBSP = " ";
+    const NBSP = "\xC2\xA0";
 
     /**
      * teardown / cleanup
--- a/tests/Zend/Locale/FormatTest.php
+++ b/tests/Zend/Locale/FormatTest.php
@@ -940,9 +940,9 @@ class Zend_Locale_FormatTest extends PHPUnit_Framework_TestCase
      */
     public function testToFloatSetlocale()
     {
-        setlocale(LC_ALL, 'fr_FR@euro'); // test setup
+        $locale = setlocale(LC_ALL, 0);
+        setlocale(LC_ALL, 'fr_FR@euro');
 
-        //var_dump( setlocale(LC_NUMERIC, '0')); // this is the specific setting of interest
         $locale_fr = new Zend_Locale('fr_FR');
         $locale_en = new Zend_Locale('en_US');
         $params_fr = array('precision' => 2, 'locale' => $locale_fr);
@@ -950,9 +950,11 @@ class Zend_Locale_FormatTest extends PHPUnit_Framework_TestCase
         $myFloat = 1234.5;
         $test1 = Zend_Locale_Format::toFloat($myFloat, $params_fr);
         $test2 = Zend_Locale_Format::toFloat($myFloat, $params_en);
+
+        setlocale(LC_ALL, $locale);
+
         $this->assertEquals("1" . self::NBSP . "234,50", $test1);
         $this->assertEquals("1,234.50", $test2);
-        // placing tearDown here (i.e. restoring locale) won't work, if test already failed/aborted above.
     }
 
     /**
@@ -1114,10 +1116,13 @@ class Zend_Locale_FormatTest extends PHPUnit_Framework_TestCase
     public function testCheckDateFormatDoesNotEmitNoticeWhenNoOptionsAreNotProvided()
     {
         try {
+            $locale = setlocale(LC_ALL, 0); // read current locale
             setlocale(LC_ALL, 'en_US'); // test setup
             Zend_Locale_Format::setOptions(array('date_format' => 'yyyy-MM-dd'));
+            $checkDateFormat = Zend_Locale_Format::checkDateFormat('2011-10-21', array());
+            setlocale(LC_ALL, $locale); // restore previous locale
 
-            $this->assertTrue(Zend_Locale_Format::checkDateFormat('2011-10-21', array()));
+            $this->assertTrue($checkDateFormat);
         } catch ( PHPUnit_Framework_Error_Notice $ex ) {
             $this->fail('Zend_Locale_Format::checkDateFormat emitted unexpected E_NOTICE');
         }
--- a/tests/Zend/Locale/FormatTest.php
+++ b/tests/Zend/Locale/FormatTest.php
@@ -941,20 +941,28 @@ class Zend_Locale_FormatTest extends PHPUnit_Framework_TestCase
     public function testToFloatSetlocale()
     {
         $locale = setlocale(LC_ALL, 0);
-        setlocale(LC_ALL, 'fr_FR@euro');
+        try {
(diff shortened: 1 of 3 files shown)


glensc:
https://github.com/zf1s/zf1/pull/49 has been merged, so can rebase now to see that pdo tests will start to fail because they are not locale aware.

glensc:
I'm unable to reproduce the "locales break float and create sql injection' problem.
This should have been the job that failed, but it didn't:
- https://github.com/zf1s/zf1/runs/2112111083

but I guess it's not that important anymore, as the code passes now.

tests/Zend/Locale/FormatTest.php
```diff
@@ -35,6 +35,12 @@
  */
 class Zend_Locale_FormatTest extends PHPUnit_Framework_TestCase
 {
+    /**
+     * Constant for Non-breaking space UTF-8 encoded value.
+     * https://en.wikipedia.org/wiki/Non-breaking_space
+     */
+    const NBSP = " ";
```
falkenhawk, line 42: Perhaps this would be safer and more explicit? 🤔 (not sure if it will work)
```suggestion
    const NBSP = "\u{00A0}";
```

tests/Zend/Validate/FloatTest.php
```diff
@@ -35,12 +35,22 @@
  */
 class Zend_Validate_FloatTest extends PHPUnit_Framework_TestCase
 {
+    /**
+     * Constant for Non-breaking space UTF-8 encoded value.
+     * https://en.wikipedia.org/wiki/Non-breaking_space
+     */
+    const NBSP = " ";
```
falkenhawk, line 42: 🤔
```suggestion
    const NBSP = "\u{00A0}";
```

tests/Zend/Validate/FloatTest.php
```diff
@@ -35,12 +35,22 @@
...
+
     /**
      * Zend_Validate_Float object
      *
      * @var Zend_Validate_Float
      */
     protected $_validator;
+    /**
```
falkenhawk, line 50: just a newline between docblocked vars
```suggestion

    /**
```

tests/Zend/Locale/FormatTest.php
```diff
@@ -1108,10 +1113,12 @@ public function testToNumberWithoutFormatWithPrecision()
     public function testCheckDateFormatDoesNotEmitNoticeWhenNoOptionsAreNotProvided()
     {
         try {
-            setlocale(LC_ALL, 'en_US'); // test setup
+            $locale = setlocale(LC_ALL, 'en_US'); // test setup
             Zend_Locale_Format::setOptions(array('date_format' => 'yyyy-MM-dd'));
+            $checkDateFormat = Zend_Locale_Format::checkDateFormat('2011-10-21', array());
+            setlocale(LC_ALL, $locale);
```
falkenhawk, line 1119: is `$locale` meant to save the previous value? As per https://www.php.net/manual/en/function.setlocale.php#refsect1-function.setlocale-returnvalues `setlocale` returns "new current value". 

Maybe this trick should be used instead to retrieve the locale before the change https://stackoverflow.com/a/33025911/3729316
```suggestion
            $locale = setlocale(LC_ALL, 0); // read current locale
            setlocale(LC_ALL, 'en_US'); // test setup
            Zend_Locale_Format::setOptions(array('date_format' => 'yyyy-MM-dd'));
            $checkDateFormat = Zend_Locale_Format::checkDateFormat('2011-10-21', array());
            setlocale(LC_ALL, $locale); // restore previous locale
```

tests/Zend/Locale/FormatTest.php
glensc: unicode literals require php 7.0

```
➔ php53 -r 'var_dump("\u{00A0}");'
string(8) "\u{00A0}"
➔ php56 -r 'var_dump("\u{00A0}");'
string(8) "\u{00A0}"
➔ php70 -r 'var_dump("\u{00A0}");'
Command line code:1:
string(2) " "
```

glensc: Safe, was to introduce a constant :)

tests/Zend/Locale/FormatTest.php
glensc: `setlocale(LC_ALL, 0);` must be used to retrieve the previous value, maybe I missed some code. or messed up while rebasing.

tests/Zend/Locale/FormatTest.php
falkenhawk: Ok... perhaps `"\xE2\x80\x8A"`? This should be compatible with php5...
I am not a fan of using invisible/indistinguishable special unicode characters in places where they can get lost over time in an accidental change of file encoding 🙊

glensc: hmm, wrong byte order?

```
➔ php70 -r 'var_dump([$a="\u{00A0}", $b="\xE2\x80\x8A", "\u{00A0}" === "\xE2\x80\x8A"]);'
Command line code:1:
array(3) {
  [0] =>
  string(2) " "
  [1] =>
  string(3) " "
  [2] =>
  bool(false)
}
```

I'll try to find the byte value on my own too, maybe you messed up :)

glensc:
```
➔ php70 -r 'var_dump("\u{00A0}" === "\xC2\xA0");'
Command line code:1:
bool(true)
```

falkenhawk: I just shamelessly copied it from https://stackoverflow.com/a/48554122/3729316 without checking 😅

tests/Zend/Locale/FormatTest.php
```diff
@@ -1108,10 +1116,13 @@ public function testToNumberWithoutFormatWithPrecision()
     public function testCheckDateFormatDoesNotEmitNoticeWhenNoOptionsAreNotProvided()
     {
         try {
+            $locale = setlocale(LC_ALL, 0); // read current locale
             setlocale(LC_ALL, 'en_US'); // test setup
             Zend_Locale_Format::setOptions(array('date_format' => 'yyyy-MM-dd'));
+            $checkDateFormat = Zend_Locale_Format::checkDateFormat('2011-10-21', array());
+            setlocale(LC_ALL, $locale); // restore previous locale
```
falkenhawk, line 1123: should this also be copied to the `catch` block below? 🤔

falkenhawk (approved):
I'm very glad that you sorted this issue out

tests/Zend/Locale/FormatTest.php
glensc: err? what? why?

falkenhawk: I thought of a case when `Zend_Locale_Format::checkDateFormat` would throw a notice, caught by `} catch ( PHPUnit_Framework_Error_Notice $ex ) {` below... then the locale would not be reverted

glensc: oh, wow. opened pandora box here. looks like the tests are incomplete and inconsistent.

1) `Zend_LocaleTest` backs up `LC_ALL` in `setUp`:
- https://github.com/zf1s/zf1/blob/8e1040b4a5c147f8751994a597253d9c460b10cd/tests/Zend/LocaleTest.php#L62

and in `tearDown` restores after each test:
- https://github.com/zf1s/zf1/blob/8e1040b4a5c147f8751994a597253d9c460b10cd/tests/Zend/LocaleTest.php#L80-L89


afaik phpunit calls tearDown after each test regardless of the test outcome. so wrapping and restoring locales in LocaleTest regarding `LC_ALL` would not be needed. but any other `LC_*` variable that is altered, need to be restored separately. even LC_ALL overrides all other variables, they still have effect if LC_ALL is unset, so need to restore.

also must note that phpunit 3.7 does not recover from subclasses of `\Error` and `\Throwable`.
ideally, no method should throw those new exceptions if all input is validated :D

2) `Zend_Locale_FormatTest` restores locales if TESTS_ZEND_LOCALE_FORMAT_SETLOCALE is set

- https://github.com/zf1s/zf1/blob/8e1040b4a5c147f8751994a597253d9c460b10cd/tests/Zend/Locale/FormatTest.php#L43-L52

it restores LC_ALL to the value of `TESTS_ZEND_LOCALE_FORMAT_SETLOCALE`. we do not set `TESTS_ZEND_LOCALE_FORMAT_SETLOCALE`.

also, the `Zend_Locale_FormatTest` contains a test that alters other locale settings than `LC_ALL`, so these must be restored.

falkenhawk (approved):
@glensc Is it safe to merge now? It looks good to me.

glensc:
@falkenhawk do not remember. merge if it's ok with you.

tests/Zend/Locale/FormatTest.php
falkenhawk: Thank you for detailed info. If the case you mentioned in 2) still makes troubles we may return to it later.

PR #48, opened 2021-01-26
Rename tests/TestConfiguration.php.dist to TestConfiguration.dist.php

glensc:
Having native extension to files, enables proper use of IDE static inspection


falkenhawk (approved):
Nice!

PR #42, opened 2020-12-31
Remove usage of $php_errormsg

Megatherium:
The track_errors directive and the global $php_errormsg variable have been removed in PHP8. The same functionality can be achieved via error_get_last() thereby making the framework compatible with PHP8 in this respect.

closes #35

PR #52, opened 2021-02-05
Enable "fail-fast" env for setup-php

glensc:
This is to ensure if an extension setup failed, the job is stopped.

See https://github.com/shivammathur/setup-php/issues/407#issuecomment-773675741

refs:
- https://github.com/zf1s/zf1/pull/42#issuecomment-773329658


falkenhawk (approved):
Thanks!

Issue #51, opened 2021-02-03
PHP 8.0 porting status


glensc:
List of documented backward-incompatible changes:
- https://www.php.net/manual/en/migration80.incompatible.php

# Tasks
- [x] PHP 8.0 compatibility patch: #32
  - [x] Fix a bug in Zend_Date for PHP 8.0 #62
  - [x] Missing default values caused errorHandlerIgnore to fail under PHP 8.0, #63
  - [x] Guard file_exists call to prevent fatal error, #64
  - [x] Remove libxml deprecations for php 8.0, #65
  - [x] Do not pass &$parts as reference to resolvePharParentPath, #67
  - [x] Ensure non-negative values for bcmath, #68
  - [x] Add support for gd resource type change in php 8.0, #69
  - [x] Add guard on fclose in Zend_Mail_Storage_Mbox, #70
  - [x] Implement T_NAME_QUALIFIED token in ClassFileLocation for php 8.0, #71
  - [x] Remove openssl deprecation for php 8.0, #73
  - [x] php 8.0: Change call in ClassMapAutoloader to preserve original behaviour, #74
  - [x] Make Zend_Validate_Date work as expected under PHP 8.0, #75
  - [x] Fixed reflection deprecations for php 8.0, #76
  - [x] Small fixes for tests in favor of PHP 8.0 compat, #77
  - [x] Remove default values from method signatures, #78
  - [x] Fix vsprintf TypeError for php 8.0, #79
  - [x] Enforce types for PHP 8.0, #80
  - [x] Fix Zend_Form_ElementTest, #81
  - [x] Fix iterators usage, #82
  - [x] DB2 constants are missing when setting up db factory, #83
  - [x] Catch ValueError on PHP 8.0 for Pdo Statement #85
- [x] Remove usage of $php_errormsg: #35, #42
- [x] version_compare does not support `gte`: https://github.com/zf1s/zf1/issues/51#issuecomment-772469392, https://github.com/zf1s/zf1/pull/26
- [x] iconv_substr behaves differently under PHP 8: #31
- [x] Zend_Log: catch TypeError/ValueError for PHP 8.0, #88
- [x] Fix error handler for PHP8 usage #99
- [x] Fix Zend_Db issues with php8 #106
- [x] php8 Zend_Log compatibility fixes #107
- [x] [zend-paginator] prevent fatal error on php8 #109
- [x] [zend-queue] prevent TypeError on md5 of non-string message #110 
[...]

PR #26, opened 2020-09-28
Zend XML - fix version comparison operator

partikus:
This PR fixes an invalid operator. **gte** is not supported.

_If the third optional operator argument is specified, test for a particular relationship. The possible operators are: <, lt, <=, le, >, gt, >=, ge, ==, =, eq, !=, <>, ne respectively._

https://www.php.net/manual/en/function.version-compare.php

PR #53, opened 2021-02-05
Replace version_compare on PHP_VERSION with PHP_VERSION_ID check

glensc:
> As described in https://bugs.php.net/bug.php?id=80503 
> 
> PHP 8 doesn't support "gte' operator for version_compare() 
> 
> I have found this one triggered only in fpm mode which break several things like Zend_Currency
> https://github.com/Shardj/zf1-future/blob/c84dedf0e948d9ccaa7a2ac234cc6f6787f05d45/library/Zend/Xml/Security.php#L172
> 
> EDIT: The above problem is fixed in #26 already (1.13.4 release)

Refs:
- https://github.com/zf1s/zf1/issues/51#issuecomment-772469392
- https://github.com/Shardj/zf1-future/issues/128#issue-800217669

### DIFF
--- a/packages/zend-db/library/Zend/Db/Adapter/Pdo/Mysql.php
+++ b/packages/zend-db/library/Zend/Db/Adapter/Pdo/Mysql.php
@@ -102,7 +102,7 @@ class Zend_Db_Adapter_Pdo_Mysql extends Zend_Db_Adapter_Pdo_Abstract
         }
 
         if (!empty($this->_config['charset'])
-            && version_compare(PHP_VERSION, '5.3.6', '<')
+            && PHP_VERSION_ID < 50306
         ) {
             $initCommand = "SET NAMES '" . $this->_config['charset'] . "'";
             $this->_config['driver_options'][1002] = $initCommand; // 1002 = PDO::MYSQL_ATTR_INIT_COMMAND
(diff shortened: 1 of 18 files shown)


packages/zend-exception/library/Zend/Exception.php
```diff
@@ -42,7 +42,7 @@ class Zend_Exception extends Exception
      */
     public function __construct($msg = '', $code = 0, Exception $previous = null)
     {
-        if (version_compare(PHP_VERSION, '5.3.0', '<')) {
+        if (PHP_VERSION_ID < 50300) {
```
glensc, line 45: There are some code that checks for php 5.3:

```php
        if (PHP_VERSION_ID < 50300) {
            parent::__construct($msg, (int) $code);
            $this->_previous = $previous;
        } else {
            parent::__construct($msg, (int) $code, $previous);
        }
```

but as we require only 5.3, perhaps drop such low version checks?

falkenhawk: go ahead and clean it up if you care (and have time for it) please :)

glensc: Will have to go manually over each check anyway, as could not find automation.

glensc: the code part is done and ready for merge:
- https://github.com/zf1s/zf1/pull/54

tests part needs more effort, as there are a lot of checks, so created as separate PR:
- https://github.com/zf1s/zf1/pull/55

glensc:
All "rc" version checks were rounded down for simplicity as pointless to support that gradual difference.

falkenhawk (approved):
it's fine to round down that single `rc` check - no need to go further with checking `PHP_EXTRA_VERSION` or something

PR #54, opened 2021-02-05
Drop code supporting php older than 5.3.3

glensc:
This drops code lower than the version specified in composer.json, currently PHP 5.3.3.

Dependency for https://github.com/zf1s/zf1/pull/53

See: https://github.com/zf1s/zf1/pull/53#discussion_r570932201

PR #55, opened 2021-02-05
Drop tests covering php older than 5.3.3

glensc:
Moved out of https://github.com/zf1s/zf1/pull/54

Dependency for https://github.com/zf1s/zf1/pull/53

See: https://github.com/zf1s/zf1/pull/53#discussion_r570932201


falkenhawk (approved):
👏

PR #56, opened 2021-02-06
Do not check get_magic_quotes_gpc value for php 5.4+

glensc:
Fixes https://github.com/Shardj/zf1-future/issues/129


glensc:
@falkenhawk perhaps a patch release when this is merged. should not hold off too long with releases when changes are merged.

falkenhawk (approved):
Great catch! thanks.

PR #58, opened 2021-02-08
Allow newer php-parallel-lint/php-parallel-lint for php 8.0

glensc:
The problem is that 5.3 can only use 1.0.0 as newer versions require newer runtime.

And 8.0.0 can't use 1.0.0 due to curly braces deprecation:

> `PHP Fatal error:  Array and string offset access syntax with curly braces is no longer supported`

Example:
- https://github.com/glensc/php-zf1s/runs/1855558100?check_suite_focus=true

Parts of the commit originally by @Megatherium

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "symfony/polyfill-php70": "^1.19"
     },
     "require-dev": {
-        "php-parallel-lint/php-parallel-lint": "1.0.0",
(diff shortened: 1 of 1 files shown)


falkenhawk (approved):

glensc:
Rebased and dependant PR's ready for review/merge.

PR #59, opened 2021-02-08
Add php 8.0 to GitHub Actions ignoring its errors

glensc:
The example given in the documentation:
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations

does not actually work as expected:
- https://github.com/actions/toolkit/issues/399

So, this uses the same "experimental: true" setup, but uses the `FAILURE_ACTION` env variable to skip errors from the experimental PHP configuration (8.0 for now).

cc @falkenhawk

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,6 +11,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        experimental:
+          - false
         php:
           - "5.3"
           - "5.4"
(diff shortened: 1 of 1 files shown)


glensc:
This requires https://github.com/zf1s/zf1/pull/58 to be merged first, as the ignore error is applied only to the phpunit run.

falkenhawk:
We should remove the `experimental` flag in #32 then, right?

glensc:
> We should remove the `experimental` flag in #32 then, right?

For 8.0 yes, but I would keep the "framework" in place for 8.1, or perhaps for 5.4/5.5 as those PHP builds are currently problematic.

falkenhawk (approved):

PR #61, opened 2021-03-01
Fix Float validator that doesn't know how to handle an int

glensc:
Ports fix from https://github.com/Shardj/zf1-future/issues/135

Author comments (from https://github.com/Shardj/zf1-future/issues/135#issuecomment-788133591):
> Looks like whoever wrote Float.php didn't pay much attention to ints.

### DIFF
--- a/packages/zend-validate/library/Zend/Validate/Float.php
+++ b/packages/zend-validate/library/Zend/Validate/Float.php
@@ -105,7 +105,7 @@ class Zend_Validate_Float extends Zend_Validate_Abstract
      *
      * Returns true if and only if $value is a floating-point value
      *
-     * @param  string $value
+     * @param mixed $value
      * @return boolean
      */
     public function isValid($value)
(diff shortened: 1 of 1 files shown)


glensc:
@falkenhawk please validate the change before merging, I only copied the change over from the other repo to keep the repos differences down.

michaelcaldwell-st (approved):

glensc:
Tests fail now, so this is apparently not complete, back to Draft:


```
There were 2 failures:

1) Zend_Validate_FloatTest::testBasic
Failed asserting that false matches expected true.

/home/runner/work/zf1/zf1/tests/Zend/Validate/FloatTest.php:92

2) Zend_Validate_FloatTest::testNoZendLocaleButPhpLocale
Failed asserting that false is true.

/home/runner/work/zf1/zf1/tests/Zend/Validate/FloatTest.php:141

```

michaelcaldwell-st:
@glensc any update on this?  What are the assertions that are failing?  Are they dependent on actually incorrect behavior?  This fix would still be helpful to us.  Thanks!

falkenhawk:
@glensc I believe this has been fixed already in #16. Same story as https://github.com/zf1s/zf1/pull/100#issuecomment-916844626

PR #63, opened 2021-03-13
Missing default values caused errorHandlerIgnore to fail under PHP 8.0

glensc:
The changed tests all change their errorhandlers in some tests. As the
last param ($errContext) had no default value assigned the tests would
error out.

Extracted from #32


falkenhawk (approved):
✔

PR #65, opened 2021-03-13
Remove libxml deprecations for php 8.0

glensc:
On libxml >= 2.9 the entity loader is disabled by default. Calls to
libxml_disable_entity_loader are now performed conditionally with lower
versions.

Extracted from #32


falkenhawk (approved):
💚

PR #67, opened 2021-03-13
Do not pass &$parts as reference to resolvePharParentPath

glensc:
Extracted change made by @Megatherium from https://github.com/zf1s/zf1/pull/32

### DIFF
--- a/packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
+++ b/packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
@@ -233,7 +233,7 @@ class Zend_Loader_ClassMapAutoloader implements Zend_Loader_SplAutoloader
      * @param  array $parts 
      * @return void
      */
-    public static function resolvePharParentPath($value, $key, &$parts)
+    public static function resolvePharParentPath($value, $key, $parts)
     {
         if ($value !== '...') {
             return;
--- a/packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
+++ b/packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
@@ -208,7 +208,10 @@ class Zend_Loader_ClassMapAutoloader implements Zend_Loader_SplAutoloader
         $prependSlash = $parts && $parts[0] === '' ? '/' : '';
         $parts = array_values(array_filter($parts, array(__CLASS__, 'concatPharParts')));
 
-        array_walk($parts, array(__CLASS__, 'resolvePharParentPath'), $parts);
+        foreach ($parts as $key => $value) {
+            self::resolvePharParentPath($value, $key, $parts);
+        }
+
         if (file_exists($realPath = 'phar://' . $prependSlash . implode('/', $parts))) {
             return $realPath;
         }
--- a/tests/Zend/Loader/ClassMapAutoloaderTest.php
+++ b/tests/Zend/Loader/ClassMapAutoloaderTest.php
@@ -36,6 +36,11 @@ if (!defined('PHPUnit_MAIN_METHOD')) {
  */
 class Zend_Loader_ClassMapAutoloaderTest extends PHPUnit_Framework_TestCase
 {
+    /**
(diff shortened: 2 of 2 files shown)


staabm:
offtopic question: I can see lot of PRs from your side, which is awesome.

it seems you are planning to add php8 support to zf1?
(asking for a friend, because we also have some apps running zf1 ;-))

glensc:
@staabm yes. it's the details all in mr body. click the link there.

glensc:
php 5.4 flakyness:

```

There was 1 error:

1) Zend_Cache_FileBackendTest::testSaveWithNullLifeTime
Zend_Cache_Exception: cache_dir "/home/runner/work/zf1/zf1/tests/Zend/Cache/zend_cache_tmp_dir_031421030011/" must be a directory

/home/runner/work/zf1/zf1/packages/zend-cache/library/Zend/Cache.php:209
/home/runner/work/zf1/zf1/packages/zend-cache/library/Zend/Cache/Backend/File.php:177
/home/runner/work/zf1/zf1/packages/zend-cache/library/Zend/Cache/Backend/File.php:128
/home/runner/work/zf1/zf1/tests/Zend/Cache/FileBackendTest.php:65

```

packages/zend-loader/library/Zend/Loader/ClassMapAutoloader.php
```diff
@@ -233,7 +233,7 @@ public static function concatPharParts($part)
      * @param  array $parts 
      * @return void
      */
-    public static function resolvePharParentPath($value, $key, &$parts)
+    public static function resolvePharParentPath($value, $key, $parts)
     {
         if ($value !== '...') {
```
falkenhawk, line 238: I looked at this method and the logic does not make sense to me. The purpose of it was to remove current and previous segment in a path, where current segment was `...`? Is triple-dots even ever used in phar paths? Moreover this method would be buggy if `...` was at the beginning of a path.

If the original intention was to have `..` there instead of `...` then it would maybe make more sense... but the logic here to mutate `$parts` array passed by reference looks awkward and is forbidden in php8 while using `array_walk`.

It looks like a workaround for a super-rare edge case to me. The case with `...` is also not tested. We could try to refactor it in a loop or something, but I think I'd get rid of this method completely.

glensc: @falkenhawk but ths pr did not modify line including `if ($value !== '...') {`

falkenhawk: Yes, it removed reference from `$parts` which made the method obsolete, but I tried to figure out the original purpose of the method...

falkenhawk: Just noticed this https://github.com/zf1s/zf1/pull/74 - it actually addresses the same issue.
But, I still wonder, why there is a condition for `...` inside `resolvePharParentPath` - it just looks off to me.

Btw, when I change that to `..` in #74 to actually allow the `$parts` array to be modified in tests, ...it crashes in tests for me locally on php 7.3 (win) with a weird error.

```
composer test tests/Zend/Loader/ClassMapAutoloaderTest.php
> phpunit "tests/Zend/Loader/ClassMapAutoloaderTest.php"
PHPUnit 3.7.38 by Sebastian Bergmann.

Configuration read from D:\projects\ovosdev\zf1s\phpunit.xml.dist

...........Script phpunit handling the test event returned with error code -1073741819
```

Per https://www.php.net/manual/en/function.array-walk.php :
> Only the values of the array may potentially be changed; its structure cannot be altered, i.e., the programmer cannot add, unset or reorder elements. If the callback does not respect this requirement, the behavior of this function is undefined, and unpredictable. 

The logic here is definitely not very safe 🙈

glensc: I've moved everything but this change to separate PRs, so they won't suffer the delay by code review.

falkenhawk:
superseded by #111

PR #68, opened 2021-03-13
Ensure non-negative values for bcmath

glensc:
Beginning with PHP8 supplying a negative value as an argument for scale
in the `bcmath` functions will generate a ValueError. Negative values
supplied to the Zend math functions are ignored and 0 is used generating
the same results as before.


falkenhawk (approved):
👍

PR #69, opened 2021-03-13
Add support for gd resource type change in php 8.0

glensc:
GD function no longer returns a resource of type gd but GdImage objects
which needed to be implemented accordingly.


falkenhawk (approved):
👏

PR #71, opened 2021-03-13
Implement T_NAME_QUALIFIED  token in ClassFileLocation for php 8.0

glensc:
T_NAME_QUALIFIED has been added with PHP8 and needed to be handled


falkenhawk (approved):
✅

PR #73, opened 2021-03-13
Remove openssl deprecation for php 8.0

glensc:
openssl_free_key has been deprecated in PHP 8.0 and is called
automatically. The explicit calls are still performed conditionally in
older PHP versions.


falkenhawk (approved):
Thanks!

PR #74, opened 2021-03-13
php 8.0: Change call in ClassMapAutoloader to preserve original behaviour

glensc:
In PHP8 the third param of array_walk ($userdata) is automatically
passed by value thus calling a function that expects pass by reference
as the third param is not possible.
Simply changing the array_walk to a foreach loop allows us to fulfill
the pass by reference signature.


glensc:
This branch commit was moved into https://github.com/zf1s/zf1/pull/67

PR #77, opened 2021-03-13
Small fixes for tests in favor of PHP 8.0 compat

glensc:

### DIFF
--- a/tests/Zend/Db/Statement/TestCommon.php
+++ b/tests/Zend/Db/Statement/TestCommon.php
@@ -329,6 +329,10 @@ abstract class Zend_Db_Statement_TestCommon extends Zend_Db_TestSetup
             $this->assertTrue($e instanceof Zend_Db_Statement_Exception,
                 'Expecting object of type Zend_Db_Statement_Exception, got '.get_class($e));
             $this->assertRegExp('#invalid fetch mode#i', $e->getMessage());
+        } catch (Throwable $e) {
+            $this->assertTrue($e instanceof ValueError,
+                'Expecting object of type Zend_Db_Statement_Exception, got '.get_class($e));
+            $this->assertRegExp('#must be a bitmask#i', $e->getMessage());
         }
     }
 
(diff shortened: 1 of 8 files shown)


glensc:
The failing test for php 5.4:

```
PHP Catchable fatal error:  Argument 1 passed to Zend_Log_Writer_Abstract::setFormatter() must implement interface Zend_Log_Formatter_Interface, instance of stdClass given, called in /home/runner/work/zf1/zf1/tests/Zend/Log/Writer/AbstractTest.php on line 68 and defined in /home/runner/work/zf1/zf1/packages/zend-log/library/Zend/Log/Writer/Abstract.php on line 97
...................
Error: Process completed with exit code 1.
```

- https://github.com/zf1s/zf1/runs/2102768787

is either flaky or the change to resolve this is in some other PR.

glensc:
The logFormatter was also discussed here:
- https://github.com/zf1s/zf1/pull/32#issuecomment-775157286

tests/Zend/Filter/HtmlEntitiesTest.php
```diff
@@ -214,7 +214,7 @@ public function testQuoteStyleQuotesEncodeNone()
      */
     public function testCorrectsForEncodingMismatch()
     {
-        if (PHP_VERSION_ID >= 50400) {
+        if (PHP_VERSION_ID >= 70000 && PHP_VERSION_ID < 70100) {
```
falkenhawk, line 217: is there any chance that changes in this file could affect the failing Logger error on PHP 5.4? 🙈

tests/Zend/Db/Statement/TestCommon.php
```diff
@@ -329,6 +329,10 @@ public function testStatementSetFetchModeInvalidException()
             $this->assertTrue($e instanceof Zend_Db_Statement_Exception,
                 'Expecting object of type Zend_Db_Statement_Exception, got '.get_class($e));
             $this->assertRegExp('#invalid fetch mode#i', $e->getMessage());
+        } catch (Throwable $e) {
+            $this->assertTrue($e instanceof ValueError,
+                'Expecting object of type Zend_Db_Statement_Exception, got '.get_class($e));
+            $this->assertRegExp('#must be a bitmask#i', $e->getMessage());
```
glensc, line 335: This and other Throwable, ValueError fixes to tests are wrong in this PR. Code must be adjusted to keep previous behavior, even if it's wrong behavior.

For example, outer code expects Zend_Db_Statement_Exception to be thrown, but instead, ValueError is thrown, and the whole application breaks. Even old PHPUnit breaks because \Error is not a subclass of \Exception.

My version of the problem fix:
- https://github.com/zf1s/zf1/pull/85

falkenhawk:
These adjustments are no longer necessary. Thanks!

PR #78, opened 2021-03-15
Remove default values from method signatures

glensc:
As optional before required params are forbidden in PHP 8.0 these signatures
needed to be changed and the functionality of the method adapted
accordingly to still retain the original behavior.

Moved out of https://github.com/zf1s/zf1/pull/67, 
which was moved out from https://github.com/zf1s/zf1/pull/32

### DIFF
--- a/packages/zend-pdf/library/Zend/Pdf/Element/Reference.php
+++ b/packages/zend-pdf/library/Zend/Pdf/Element/Reference.php
@@ -88,7 +88,7 @@ class Zend_Pdf_Element_Reference extends Zend_Pdf_Element
      * @param Zend_Pdf_ElementFactory $factory
      * @throws Zend_Pdf_Exception
      */
-    public function __construct($objNum, $genNum = 0, Zend_Pdf_Element_Reference_Context $context, Zend_Pdf_ElementFactory $factory)
+    public function __construct($objNum, $genNum, Zend_Pdf_Element_Reference_Context $context, Zend_Pdf_ElementFactory $factory)
     {
         if ( !(is_integer($objNum) && $objNum > 0) ) {
             // require_once 'Zend/Pdf/Exception.php';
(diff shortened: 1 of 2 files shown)


glensc:
I would have personally just filled defaults to the rest of the arguments as `null`.

packages/zend-pdf/library/Zend/Pdf/Element/Reference.php
```diff
@@ -88,8 +88,12 @@ class Zend_Pdf_Element_Reference extends Zend_Pdf_Element
...
      * @throws Zend_Pdf_Exception
      */
-    public function __construct($objNum, $genNum = 0, Zend_Pdf_Element_Reference_Context $context, Zend_Pdf_ElementFactory $factory)
+    public function __construct($objNum, $genNum, Zend_Pdf_Element_Reference_Context $context, Zend_Pdf_ElementFactory $factory)
     {
+        if (empty($genNum)) {
+            $genNum = 0;
+        }
+
         if ( !(is_integer($objNum) && $objNum > 0) ) {
```
falkenhawk, line 97:
```suggestion
        if ( !(is_integer($objNum) && $objNum > 0) ) {
```

packages/zend-service-windowsazure/library/Zend/Service/WindowsAzure/Storage/Queue.php
```diff
@@ -519,9 +519,9 @@ public function clearMessages($queueName = '')
 	 * @param Zend_Service_WindowsAzure_Storage_QueueMessage $message Message to delete from queue. A message retrieved using "peekMessages" can NOT be deleted!
 	 * @throws Zend_Service_WindowsAzure_Exception
 	 */
-	public function deleteMessage($queueName = '', Zend_Service_WindowsAzure_Storage_QueueMessage $message)
+	public function deleteMessage($queueName, Zend_Service_WindowsAzure_Storage_QueueMessage $message)
 	{
-		if ($queueName === '') {
+		if (empty($queueName)) {
```
falkenhawk, line 524:
```suggestion
		if ($queueName === '') {
```

falkenhawk (commented):
Looking at the code, the rest of arguments are required and they can't be null for the components to work properly.

Since values of `$genNum` and `$queueName` are validated anyway, I'd just remove the defaults and not change any logic inside method bodies. Would that be okay with you?

I suspect the default values on those arguments could be just unintended leftovers. 🤷‍♂️

falkenhawk (approved):
Is it okay to merge this?

glensc:
> Is it okay to merge this?

@falkenhawk you requested changes, I have not reviewed these, so marked back as Draft.

glensc:
Rebased and applied proposed fixes. Removed Draft

PR #80, opened 2021-03-15
Enforce types for PHP 8.0

glensc:
PHP 8.0 is a stickler for types necessitating some explicit casts.

Moved out of https://github.com/zf1s/zf1/pull/67, which was moved out from https://github.com/zf1s/zf1/pull/32


falkenhawk (approved):
🧹

PR #81, opened 2021-03-15
Fix Zend_Form_ElementTest

glensc:
@Megatherium on 3 Dec 2020:

> The behaviour hasn't really changed because the intent of the test is to
> use an array of arrays in setOptions. What has changed is that PHP 8.0
> supports calling functions with named parameters and will throw a Fatal
> if you try to access a parameter by name that does not exist.
> Digits::__construct has no param named bar ergo we get a Fatal.
> 

See discussion here:
- https://github.com/zf1s/zf1/pull/32#discussion_r534991212

Extracted out of https://github.com/zf1s/zf1/pull/77 which was @Megatherium changes from https://github.com/zf1s/zf1/pull/32

### DIFF
--- a/tests/Zend/Form/ElementTest.php
+++ b/tests/Zend/Form/ElementTest.php
@@ -48,6 +48,11 @@ if (!defined('PHPUnit_MAIN_METHOD')) {
  */
 class Zend_Form_ElementTest extends PHPUnit_Framework_TestCase
 {
+    /**
+     * @var Zend_Form_Element
+     */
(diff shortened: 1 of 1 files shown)


falkenhawk (approved):
Thank you for extracting this change into a separate PR.

glensc:
> Thank you for extracting this change into a separate PR.

https://github.com/zf1s/zf1/pull/77 was left as a catch-all for the remaining of @Megatherium changes.

PR #82, opened 2021-03-15
Fix iterators usage

glensc:
Made the handling of iterators in Zend_Controller_Action_HelperBroker_PriorityStackTest compatible with all versions.

From https://github.com/zf1s/zf1/pull/32#discussion_r534987951:
> @falkenhawk on 3 Dec 2020 Member
> does php8 throw on calling current() / next() on an iterator?
> 
> @Megatherium Megatherium on 3 Dec 2020 Author
> It returns false and thus get_class throws a TypeError
> 

Extracted out of https://github.com/zf1s/zf1/pull/77 which was @Megatherium changes from https://github.com/zf1s/zf1/pull/32


glensc:
sorry, I do not have context of what was the problem, maybe it was discussed in https://github.com/zf1s/zf1/pull/32

falkenhawk:
> 
> 
> sorry, I do not have context of what was the problem, maybe it was discussed in #32

must be this https://github.com/zf1s/zf1/pull/32/files#r534987951

glensc:
@falkenhawk updated MR body

falkenhawk (approved):

PR #83, opened 2021-03-15
DB2 constants are missing when setting up db factory

glensc:
In a lot of cases with PHP8 Type and ValueErrors are thrown instead of
warnings emitted. The error handling in tests needed to be modified
accordingly.

Changes for this discussion:
- https://github.com/zf1s/zf1/pull/32#discussion_r534989302

### DIFF
--- a/tests/Zend/Db/TestSetup.php
+++ b/tests/Zend/Db/TestSetup.php
@@ -80,14 +80,18 @@ abstract class Zend_Db_TestSetup extends PHPUnit_Framework_TestCase
      */
     protected function _setUpAdapter()
     {
-        $this->_db = Zend_Db::factory($this->getDriver(), $this->_util->getParams());
         try {
+            $this->_db = Zend_Db::factory($this->getDriver(), $this->_util->getParams());
             $conn = $this->_db->getConnection();
         } catch (Zend_Exception $e) {
             $this->_db = null;
             $this->assertTrue($e instanceof Zend_Db_Adapter_Exception,
                 'Expecting Zend_Db_Adapter_Exception, got ' . get_class($e));
             $this->markTestSkipped($e->getMessage());
+        } catch (Throwable $e) {
+            if (stristr($e->getMessage(), 'undefined constant')) {
+                $this->markTestSkipped();
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


tests/Zend/Db/TestSetup.php
```diff
@@ -80,14 +80,18 @@ protected function _setUpTestUtil()
...
             $conn = $this->_db->getConnection();
         } catch (Zend_Exception $e) {
             $this->_db = null;
             $this->assertTrue($e instanceof Zend_Db_Adapter_Exception,
                 'Expecting Zend_Db_Adapter_Exception, got ' . get_class($e));
             $this->markTestSkipped($e->getMessage());
+        } catch (Throwable $e) {
+            if (stristr($e->getMessage(), 'undefined constant')) {
+                $this->markTestSkipped();
+            }
```
glensc, line 94: Example:
- https://github.com/zf1s/zf1/runs/2115567186

But I'm thinking maybe this approach is wrong, better emulate in code the same behavior as older php versions.

glensc: This class can't be even loaded:
- https://github.com/zf1s/zf1/blob/1cfdc03ff89ddde66d5240c8cd920b725c946ad3/packages/zend-db/library/Zend/Db/Adapter/Db2.php#L82

The exception is thrown on class construction:
- https://github.com/zf1s/zf1/blob/1cfdc03ff89ddde66d5240c8cd920b725c946ad3/packages/zend-db/library/Zend/Db.php#L270

glensc: I think the only solution here is to assign `$_execute_mode` default value on the constructor. but no idea about is that backward compatible change or not.

glensc: 591a1ab7

falkenhawk:
In my opinion this will not be necessary as running `Zend\Db\AllTests.php` (either directly, by passing this file to phpunit command, or just letting phpunit run all tests with default configuration) also omitted other disabled adapters - effectively avoiding a fatal error with missing DB2 constant.

I am going to close this, thank you for your investigation.

PR #84, opened 2021-03-15
Use gnu parallel for validating composer.json

glensc:
On my local laptop, the runtime changes 45s to 11s.

On GitHub Actions the difference is 16s vs 10s.

And here's an example for a job where composer.json validation failed for two files:
- https://github.com/zf1s/zf1/runs/2114897482


falkenhawk (approved):
🚀

PR #85, opened 2021-03-15
Catch ValueError on PHP 8.0 for Pdo Statement

glensc:
```
PHP Fatal error:  Uncaught ValueError: PDOStatement::setFetchMode(): Argument #1 ($mode) must be a bitmask of PDO::FETCH_* constants
```

Example:
- https://github.com/zf1s/zf1/runs/2114576945

Addresses previous fix with backward-compatible behavior:
- https://github.com/zf1s/zf1/pull/77/files#r594541058

For example, outer code expects `Zend_Db_Statement_Exception` to be thrown, but instead, `\ValueError` is thrown, and the whole application breaks:
- https://github.com/zf1s/zf1/blob/1.13.4/tests/Zend/Db/Statement/TestCommon.php#L326

Note: Even old PHPUnit breaks because `\Error` is not a subclass of `\Exception` :)


falkenhawk:
superseded by #106

PR #86, opened 2021-03-15
paginator: fixed phpdoc typo

staabm:
`$this->_createPages($scrollingStyle);` returns a `stdClass` so `getPages` can never return a array as documented


glensc (approved):

falkenhawk (commented):
Could you also fix the type at https://github.com/zf1s/zf1/blob/b4542618883a1d45f14c8e5cbaa7d75d814573e5/packages/zend-paginator/library/Zend/Paginator.php#L171 while you are at it, please? 🙏

staabm:
> Could you also fix the type at

good catch. Sure thing

falkenhawk (approved):

falkenhawk:
Thank you for your contribution @staabm!

PR #87, opened 2021-03-15
fix `Zend_View_Abstract::__get()` phpdoc

staabm:
as discussed in phpstan upstream issue https://github.com/phpstan/phpstan/issues/3927#issuecomment-705760442

to access variables from within a controller via `$this->view` without a phpstan error, this phpdoc needs to be fixed.


glensc (approved):
LGTM.

but note for the future. try to explain the whole context in PR title/body, so the reviewer doesn't have to click links and figure out what did the pull request creator had in mind when they added the link.

ps: cross-linking stuff, is very okay.

falkenhawk (approved):
Thank you @staabm !

PR #90, opened 2021-05-11
Zend_Validate_Hostname: added .fun TLD domain

dec-marcin:
see: https://www.iana.org/domains/root/db/fun.html


falkenhawk:
Thank you for your contribution @dec-marcin . Do you know if there are more TLDs missing?

falkenhawk:
superseded by https://github.com/zf1s/zf1/pull/104

PR #95, opened 2021-09-10
[zend-db] Fix PHPDoc @return statement for Zend_Db_Select::query()

glensc:
Carry https://github.com/Shardj/zf1-future/pull/166

cc @rruchte


falkenhawk (approved):
Thanks!

PR #96, opened 2021-09-10
[zend-loader] Avoiding error logs when using class_exists

glensc:
Since PHP 8.0 @ doesn't mute the errors, which makes for a lot of clutter on the error log when you are using class_exists() and it doesn't.
Adding this verification solves the problem.

Carry https://github.com/Shardj/zf1-future/pull/168

cc @Dringho

### DIFF
--- a/packages/zend-loader/library/Zend/Loader.php
+++ b/packages/zend-loader/library/Zend/Loader.php
@@ -131,9 +131,13 @@ class Zend_Loader
          * Try finding for the plain filename in the include_path.
          */
         if ($once) {
-            include_once $filename;
+            if (stream_resolve_include_path($filename)) {
+                @include_once $filename;
+            }
         } else {
-            include $filename;
+            if (stream_resolve_include_path($filename)) {
+                @include $filename;
+            }
         }
 
         /**


glensc:
@falkenhawk I think we should add an additional PHP_VERSION check to fall back to the previous behavior so that older platforms performance is not degraded.

falkenhawk:
this change made tests to fail on all php versions (and fatal on 8.0) 🤔
I think I get the idea, but we didn't even had `@` suppression on `include` / `include_once` in the first place. If I remember correctly it's ensured that the loader works as expected 🤔  but maybe I am wrong, I don't remember. Do we have issues with the loader?

Dringho:
:wave:  Sorry about the mess with the commit, I'm not sure why the function won't run on PHP 7.4, according to the doc stream_resolve_include_path should be viable since PHP 5.3.2

Regardless, we could add a PHP_VERSION_ID >= 80000 check for this behavior.

Let me know your thoughts

falkenhawk:
Thanks for the comment @Dringho, no problem at all, you don't have to apologize :)

I am trying to figure out the intention and the real issue which this addressed on `zf1-future` to understand if this is also relevant for this fork.

The `@` suppression operators were added in https://github.com/Shardj/zf1-future/pull/98 ( https://github.com/Shardj/zf1-future/pull/98/files#diff-9f9e8537ebd167ad5836c179858cf7183ca65f3ede2f265db6b6564e8623c019 ) to deal with `error-logs full of PHP Warnings due to autoloads from class-exists()-checks.`

I don't think zf1s spams these kind of warnings since I remember a time when I was going through `class_exists` and Zend_Loader calls to ensure everything works both with composer autoloader and Zend Loader (where necessary). So I guess there should not be an issue, but of course my optimistic assumptions might be wrong, still I would need a test with reproduction to prove that there is something to fix on our side 😉

falkenhawk:
Once #116 lands as v1.15.0 this should not be an issue anymore.

PR #97, opened 2021-09-10
[zend-validate] Adding croatian diacritical characters to hostname validation

glensc:
Carry https://github.com/Shardj/zf1-future/pull/160

cc @kiboke


falkenhawk:
Maybe we should add the whole list instead of picking just croatian variant? 🤔
look how many there are: https://github.com/zendframework/zend-validator/blob/master/src/Hostname.php#L1651

glensc:
@falkenhawk i just wanted to sync the changes in the other repo. as the sync is not complete the picture is fuzzy. i suggest to release 1.14.0 first before any of these new pull requests.

glensc:
1.14.0 is released, so that is no longer a blocker

falkenhawk (approved):

PR #98, opened 2021-09-10
[zend-db] Fix PHPDoc typings on Zend_Db_Table_Rowset_Abstract::current

glensc:
Carry https://github.com/Shardj/zf1-future/pull/158

cc @colorninja


falkenhawk (approved):
Thank you @colorninja @glensc !

PR #100, opened 2021-09-10
[zend-validate] Ensure value is a string prior to validation

glensc:
Method expects a string value. If a float is
passed in PHP 7.4 will issue a notice for
trying to access array offset.

Carry https://github.com/Shardj/zf1-future/pull/151

cc @davejennings

### DIFF
--- a/packages/zend-validate/library/Zend/Validate/Int.php
+++ b/packages/zend-validate/library/Zend/Validate/Int.php
@@ -134,7 +134,7 @@ class Zend_Validate_Int extends Zend_Validate_Abstract
 
         } else {
             try {
-                if (!Zend_Locale_Format::isInteger($value, array('locale' => $this->_locale))) {
+                if (!Zend_Locale_Format::isInteger(strval($value), array('locale' => $this->_locale))) {
                     $this->_error(self::NOT_INT);
                     return false;
                 }


falkenhawk (commented):
could we have a test for this case please 😅

falkenhawk:
I wonder what kind of notice did the original code issue?
https://github.com/zf1s/zf1/blob/e5f2a2e736f266f5957485e177f7aa5a45a15133/packages/zend-locale/library/Zend/Locale/Format.php#L724
the first argument is not required to be `string`, it's only type-hinted in phpdoc.

Then inside `Zend_Locale_Format::isNumber()` the value is casted to string anyways.

I also saw there are cases for non-string floats + locale in https://github.com/zf1s/zf1/blob/e5f2a2e736f266f5957485e177f7aa5a45a15133/tests/Zend/Validate/IntTest.php#L61 but I could not spot any relevant notice to be thrown in a PHP 7.4 test run on master https://github.com/zf1s/zf1/runs/3565219911 🤔

davejennings:
For reference, issue I had was along the lines of this test:

```
<?php

class Zend_Validate_IntTest extends \PHPUnit\Framework\TestCase
{

    public function testNoticeOnFloatValidation()
    {
        $this->obj = new Zend_Validate_Int();
        $this->assertFalse($this->obj->isValid(0.1));
    }
}
```

Which resulted in:

```
1) Zend_Validate_IntTest::testNoticeOnFloatValidation
Trying to access array offset on value of type float

/vendor/shardj/zf1-future/library/Zend/Locale/Format.php:527
/vendor/shardj/zf1-future/library/Zend/Locale/Format.php:727
/vendor/shardj/zf1-future/library/Zend/Validate/Int.php:139
/tests/library/ZF1/Validate/Zend_Validate_IntTest.php:9
```

Quoting the float as a string worked but that's not how I interpreted the intention of the original ZF code. Going back a bit and I've had a baby since then though, so my memory isn't to be trusted.

falkenhawk:
@davejennings thank you very much for taking your time to provide more details here!

It appears we have already tackled the issue in https://github.com/zf1s/zf1/pull/16/files#diff-1782ecc91bb385acd1ea87235135268a0f47f2670f7fe63fe02b08017ef33801R510
by taking a more general approach and casting the $input var to string in `Zend_Locale_Format::isNumber()`, so I am going to close this PR.

PR #102, opened 2021-09-10
The type attribute is unnecessary for JavaScript resources

glensc:
Carries https://github.com/Shardj/zf1-future/pull/54

cc @Nazin


glensc:
Moved to https://github.com/zf1s/zf1/pull/103 due dependant changes

PR #104, opened 2021-09-10
[zend-validate] Update list of available tlds

glensc:
Based on https://data.iana.org/TLD/tlds-alpha-by-domain.txt

Carry https://github.com/Shardj/zf1-future/pull/72

cc @SanderVanLeeuwen


falkenhawk:
I tried to update the list by hand from the most recent version of https://data.iana.org/TLD/tlds-alpha-by-domain.txt, but noticed that some domains are not listed there i.a. `aigo`, `caseih`, `intel` etc. but they are still listed in http://www.iana.org/domains/root/db/ 🤔 Also, the txt list contains puny-encoded domains, whereas the web list has them listed in utf8 chars. So `$_validTlds` contains both :/

I looked up how they had it in the last zf version, and yes there was a script https://github.com/zendframework/zend-validator/blob/master/bin/update_hostname_validator.php - but it was using only the txt version but utf8 entries still existed in the `$validTlds` array of the validator so they were left there for BC probably and forgotten 🙈

The script was initially made to parse the web version in https://github.com/zendframework/zendframework/commit/da354d67ff3f9980c94f50c35ac714c206de1631 but then modified at the same day to use the txt version instead https://github.com/zendframework/zendframework/commit/3c1f8d8a2325d14675acf7a0d30654559e6b27d4#diff-40b96ff2ab5d1e0c7f28660cd8513a4a983290d6c818170453d4ccd1ad312318

falkenhawk:
I updated the list from the current txt version and decided not to remove any existing entries (which were probably copied from the web version initially) to keep backwards compatibility.

glensc:
@falkenhawk don't merge any of these "new" PR's before you make 1.14.0 release

glensc:
1.14.0 is released, so that blocker no longer applies.

falkenhawk (approved):

PR #105, opened 2021-09-10
[zend-queue] Adding support for durable subscribers and persistent message sending

glensc:
Carry https://github.com/Shardj/zf1-future/pull/144

> The ActiveMQ adapter does not support durable topic subscribers, or persistent message sending. This patch adds support for these commonly used ActiveMQ feature.

cc @rruchte


falkenhawk (approved):
Thank you! Although I would not consider this outdated lib to be a reliable mq client 😅 but people may still be using it on legacy projects and this option seems useful... - so it shouldn't hurt to add this feature

PR #106, opened 2021-09-25
Fix Zend_Db issues with php8

falkenhawk:
1) _(replaces #85)_ PHP 8.0+ throws `ValueError` on invalid arguments. Catch `ValueError` on wrong "fetch" value passed to pdo methods, transform it to expected `Zend_Db_Statement_Exception` and test for different error message php. Fixes following fatal errors:
  - `PDOStatement::setFetchMode(): Argument #1 ($mode) must be a bitmask of PDO::FETCH_* constants`
  - `PDOStatement::fetchAll(): Argument #1 ($mode) must be a bitmask of PDO::FETCH_* constants`
  - `PDOStatement::fetch(): Argument #1 ($mode) must be a bitmask of PDO::FETCH_* constants`
  
2) allow case insensitive comparison in a zend db static test
  - Prevents following test failure: `Failed asserting that 'include_once(Test/MyCompany2/Dbadapter.php): Failed to open stream: No such file or directory' contains "failed to open stream".`

3) remove wrong conditions for targeting PHP < 5.2.0 in zend-db tests. Only `minor <= 2` was checked, resulting few tests to be skipped on php 8.0 (and 7.0 & 7.1). Removed the false-path completely as php <5.3 is not supported

refs:
- https://github.com/zf1s/zf1/issues/51

### DIFF
--- a/packages/zend-db/library/Zend/Db/Statement/Pdo.php
+++ b/packages/zend-db/library/Zend/Db/Statement/Pdo.php
@@ -252,6 +252,9 @@ class Zend_Db_Statement_Pdo extends Zend_Db_Statement implements IteratorAggrega
         }
         try {
             return $this->_stmt->fetch($style, $cursor, $offset);
+        } catch (ValueError $e) {
+            // PHP 8.0+ throws ValueError on invalid arguments
+            throw new Zend_Db_Statement_Exception($e->getMessage(), $e->getCode());
         } catch (PDOException $e) {
             // require_once 'Zend/Db/Statement/Exception.php';
             throw new Zend_Db_Statement_Exception($e->getMessage(), $e->getCode(), $e);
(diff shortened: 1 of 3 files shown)


falkenhawk:
Note: even though there is still a fatal error caused by `ValueError` in another place, all (configured) zend-db tests run on php8, which can be checked at https://github.com/zf1s/zf1/runs/3707648107?check_suite_focus=true - only `Zend\Db\AllTests.php` was executed there.

Note 2: Running `Zend\Db\AllTests.php` directly with phpunit also omitted other disabled adapters - effectively avoiding a fatal error with missing DB2 constant - so we might not need fixes from #83 after all /cc @glensc

marcing (approved):

glensc:
@falkenhawk It's great that you're able to finish my started work.

PR #115, opened 2021-10-01
use mysql v5.7 in gha

falkenhawk:
instead of v5.6

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,13 +44,13 @@ jobs:
           - 11211:11211
 
       mysql:
-        image: mysql/mysql-server:5.6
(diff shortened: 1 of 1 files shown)


marcing (approved):

glensc:
please fill reasoning why? I kept the lowest version to have the most compatible tests.

falkenhawk:
simply because I believe 5.7 is more popular than 5.6 - i could not find any charts or usage statistics though.
I also similarly picked postgres v12 in #114 . I know ideally it would be to add a matrix of versions, but I did not want to invest more time to this.

falkenhawk:
Still it would be worth to test both mysql v5.x and v8.x...

PR #116, opened 2021-10-18
zend-loader and autoloader overhaul - cont.

### LINKED ISSUE #46: Zend_Translate issues 'failed to open stream' warning when using a custom adapter
Issue is on line 130 of class Zend_Translate, the 'class_exists' function is attempting to autoload a custom adapter as a 'Zend_Translate_Adapter_' and issues a warning when it is not found:

PHP Warning:  include_once(Zend/Translate/Adapter/Custom/Translate/Adapter.php): failed to open stream: No such file or directory in /var/www/comm-os/vendor/zf1s/zend-loader/library/Zend/Loader.php on line 134

Requires the 'autoload' parameter of the class_exists function to be set to false.

falkenhawk:
continued (apparently unfinished) https://github.com/zf1s/zf1/pull/1 / https://github.com/zf1s/zf1/commit/76477fbe00a198ef4376ea38c46df3960c574af8

- introduce `Zend_Loader::tryLoadClass()` to attempt loading a class but do not throw an exception when file is not found (but still throw when required class in the existing file is not defined)
- did not go for suppressing loading classes with `@` suppressor by default, since regular warnings/errors coming from a loaded file should be visible, otherwise it might be very confusing for devs
- introduce `Zend_Loader_Exception_FileNotFoundException` and `Zend_Loader_Exception_ClassNotFoundException`
  - instead of throwing `Zend_Exception` in `Zend_Loader::loadClass()` with a generic message `File \"$file\" does not exist or class \"$class\" was not found in the file`
  - `Zend_Loader_Exception_FileNotFoundException` will be thrown with message `File "$file" could not be found within configured include_path.` or
  - `Zend_Loader_Exception_ClassNotFoundException` with message `Class "$class" was not found in the file "$file".`, in their respective cases.
- since composer autoloader takes precedence and `Zend_Autoloader` / `Zend_Loader` is only an extra helper, when necessary (when still needed and configured on legacy projects) - `Zend_Loader` should not emit warnings when checking for files, by default. Added `isReadable` check inside `loadFile` before `include`/`include_once`. There might be a performance hit, but it should be okay when most of the autoloading is handled by composer autoloader anyway.
- use `stream_resolve_include_path()` in `isReadable()` - it should be more performant and the function is available since php 5.3.2
- `Zend_Loader::loadClass()` returns boolean now
- adjust combinations of `class_exists` + `Zend_Loader::loadClass()` in remaining affected places (i.a. Translate or File_Transfer or Filter adapters)
- unregister autoloader from spl_autoload when `Zend_Loader_Autoloader::resetInstance()` is called
- added more tests for `Zend_Loader` and `Zend_Autoloader`
- added tests for loading custom `Zend_Translate` adapter while having `Zend_Autoloader` enabled
- adjusted existing tests which expected "file not found" php warning to be triggered by `Zend_Loader`

should fix #46, replaces #96 and #20
Also possibly introduces a breaking change in how `Zend_Loader` works, so will probably release as v1.15

### DIFF
--- a/packages/zend-currency/library/Zend/Currency.php
+++ b/packages/zend-currency/library/Zend/Currency.php
@@ -800,10 +800,7 @@ class Zend_Currency
         if (is_string($service)) {
             // require_once 'Zend/Loader.php';
             if (!class_exists($service)) {
-                $file = str_replace('_', DIRECTORY_SEPARATOR, $service) . '.php';
-                if (Zend_Loader::isReadable($file)) {
-                    Zend_Loader::loadClass($service);
-                }
+                Zend_Loader::loadClass($service);
             }
 
             $service = new $service;
--- a/packages/zend-file-transfer/library/Zend/File/Transfer.php
+++ b/packages/zend-file-transfer/library/Zend/File/Transfer.php
@@ -67,7 +67,9 @@ class Zend_File_Transfer
         if (class_exists('Zend_File_Transfer_Adapter_' . ucfirst($adapter))) {
             $adapter = 'Zend_File_Transfer_Adapter_' . ucfirst($adapter);
         } elseif (!class_exists($adapter)) {
-            Zend_Loader::loadClass($adapter);
+            if (!Zend_Loader::tryLoadClass('Zend_File_Transfer_Adapter_' . ucfirst($adapter), null, true)) {
+                Zend_Loader::loadClass($adapter);
+            }
         }
 
         $direction = (integer) $direction;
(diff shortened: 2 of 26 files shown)


falkenhawk:
@glensc Could you take a look, please?

cc @g10-fred @saru2048 @Dringho

glensc:
Since I don't use zend-loader myself, I can't comment on this.

falkenhawk:
Internal tests on existing projects were successful. Unfortunately our usage does not cover all possible scenarios for zend-autoloader/zend-loader/composer autoloader combinations, so we are able to speak only for ourselves.

Unfortunately, we heard nothing from other users, who might be interested in this if this PR fixes their cases - for almost a year.

I will merge it now and release 1.15.0 together with other changes.

glensc:
if this has issues, they can rollback to 1.14.0 and report the issue, there aren't many other changes in 1.15.0.

PR #119, opened 2022-01-14
fix `Zend_Db_Table_Select->setIntegrityCheck()` flag-parameter phpdoc

staabm:


hc-jworman (changes requested):
Might as well fix the @var annotation on $this->_integrityCheck while you are here

glensc:
please, not vague commit messages and PR titles. I call it click-bait because without clicking to see the diff the title doesn't say anything useful.

put yourself in the context of looking over `git log` and what impression does it level if you see such commit message?

while at it. worthful reading:
- https://www.freshconsulting.com/insights/blog/atomic-commits/
- https://cbea.ms/git-commit/

glensc:
@hc-jworman your comment costs more than you making the actual change. you apparently already had the file open, make edit, submit the pr is just a few clicks away.

staabm:
fixed per review. thx for the feedback

hc-jworman:
@glensc I don't entirely agree, but I will keep that in mind in the future.

glensc:
@hc-jworman and now your reply only increased the cost of your initial comment :D now three people have had to read the comments. so, multiplied by 3, if maintainer also reads, multiplied by 4. and unlikely your comment will be actually heard (i.e someone will make the change you described). I personally didn't see the symbol you mentioned in the diff, so I stopped and didn't even look further.

falkenhawk (approved):
Thanks!

PR #126, opened 2022-06-27
[zend-session] adds "cookie_samesite" option

marcing:
adds "cookie_samesite" option - https://www.php.net/manual/en/session.configuration.php#ini.session.cookie-samesite


glensc (approved):

partikus:
@marcing please rebase after #127

partikus (approved):

Issue #57, opened 2021-02-08
publish monorepo composer package


glensc:
It seems that `zf1s/zf1` is not registered on packagist.org:

```
$ composer require zf1s/zf1
  [InvalidArgumentException]
  Could not find a matching version of package zf1s/zf1. Check the package spelling, your version constraint and that the package is available in a stability which matches your minimum-stability (stable).
```

This would help people installing "full zf1":
- https://github.com/zf1s/zf1/issues/51#issuecomment-774701307

project readme must be also updated, when this installation method works.

falkenhawk:
I am not sure if I would like to do this, I am afraid it misses the whole point of the split components and less overhead compared to checking out the whole framework.
Moreover it may require additional maintenance and qa 🤔
I am not sure if it would easily workaround the issue reported in https://github.com/zf1s/zf1/issues/51#issuecomment-774701307 if composer is not available on that project - then having the whole framework package published on packagist does not help much, does it?
Perhaps the maintainer of that project could checkout the zf1 repo directly and use it in his project? 🤔

falkenhawk:
Also if composer autoloader is not available there, then a separate, custom autoloader solution would be needed there which would load the files from individual `packages/zend-*/library/Zend` folders as there is no longer a single "root" folder which allows a PSR-0 autoloader to handle it (in contrary to https://github.com/Shardj/zf1-future/tree/master/library/Zend)

glensc:
@falkenhawk it would be a package that has all dependencies installed. like symfony has. would be a solution to developers who are ok with composer, but not with identifying which packages are needed.

falkenhawk:
After thinking about it for more that a year, we decided to stick to our past decision not to publish the whole package `zf1s/zf1` on packagist. Devs should be encouraged to install individual components, if necessary, and keep jumping off the ship in small steps 🙈 reducing dependencies to zf1 one by one.

Keep in mind, this whole thing was not meant to provide long-term support for zf1, only to allow working on legacy projects on more modern systems, while opening the possibility to migrate away from it, one component at a time.

Having everything split to single components should not be undermined by easy and lazy option to install the whole at once ;) 

Also, we do not want to bring in another official "full" package to packagist, introducing more confusion - which one to install, how would that relate to individual packages etc.

With all that said, it is still easy to install it for dev purposes / manual testing / transition period, by adding that to your `composer.json`:
```json
    "repositories": [
        {
            "type": "git",
[...]

Issue #128, opened 2022-09-20
Support for zendframework/zf1-extras


fredericgboutin-yapla:
Hi,
I'm digging my way around and I'm struggling to find any info in regards to support for "zf1-extras" or "ZendX" components like JQuery - https://framework.zend.com/manual/1.12/en/zendx.jquery.html

I found this fork - https://github.com/Shardj/zf1-extras-future - which seems "popular" so I suppose I could use it in conjunction with zf1s but I would like to hear your input on the whole matter.

Thanks!

fredericgboutin-yapla:
Well, it's been 2 weeks and I've got no feedback. 

Anyway, I went ahead and I tried to use zf1s with zf1-extras and there are 2 problems,

1. There are a lot of `require_once` in zf1-extras codebase which don't resolve with zf1s
2. zf1s is NOT a drop-in replacement for zf1, as it does NOT / cannot inherently specify in the `composer.json` a definition like,

```json
    "replace": {
        "zendframework/zendframework1": ">=1.12.20"
    }
```

All this mean that we would need to fork zf1-extras, remove the `require_once` statements, adjust the code to make it work including the unit tests, put those tests under CI for php compatibility review, etc.

Basically when someone uses zf1s it alienates itself from using other components which could provide features to zf1, for example,

- https://github.com/PHP-DI/ZF1-Bridge
- https://github.com/guilhermeblanco/zendframework1-doctrine2
- https://github.com/jokkedk/ZFDebug
- https://github.com/Shardj/zf1-extras-future

falkenhawk:
@fredericgboutin-yapla now full `zf1s/zf1` package is available. https://packagist.org/packages/zf1s/zf1 v1.15.0 - please try.

falkenhawk:
Thank you for sharing your findings @fredericgboutin-yapla !

Those two classes `Zend_Tool_Framework_Provider_DocblockManifestInterface` and `Zend_Gdata_Analytics_Goal` deserve a fix to rename/move their files so that composer autoloader does not complain.

As for php extensions listed in `require` section.. you might be right, it is probably too restrictive at the moment. Due to the nature of zf1, you may never need some of the extensions, as long as you do not use components which require them. But when you start using them, those components themselves should throw errors that required extensions are not installed.

[...]

PR #133, opened 2022-10-05
publish the whole zf1s/zf1 package

falkenhawk:
resolves https://github.com/zf1s/zf1/issues/57

despite what I wrote in https://github.com/zf1s/zf1/issues/57#issuecomment-1176125117, I've realized that it may make sense to publish the whole `zf1s/zf1` package after all, to easy up the transition period, in which devs should eventually identify and pick only those components they really use, to install (which may not ever happen, but it's ok) ;) 

Initially _(after reading through valid arguments posted by @fredericgboutin-yapla in https://github.com/zf1s/zf1/issues/128#issuecomment-1265381141 )_ I thought it would be enough if only `replace` definition `"zendframework/zendframework1": ">=1.12.20"` was added, while the whole framework could still be installed with git installation method, as described in https://github.com/zf1s/zf1/issues/57#issuecomment-1176125117,

and that
> we do not want to bring in another official "full" package to packagist, introducing more confusion - which one to install, how would that relate to individual packages etc.

but...
screw that, we do not have to be that opinionated.

I am going to install the whole package `zf1s/zf1` together with 1.15.0 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 [![Build Status](https://github.com/zf1s/zf1/actions/workflows/tests.yml/badge.svg)](https://github.com/zf1s/zf1/actions/workflows/tests.yml)
 
 This is a monorepo of a fork of Zend Framework 1, made after it's reached its EOL.
+
 All original framework's components have been split into individual packages, which can be installed separately with `composer`, e.g.
 ```
 composer require zf1s/zend-*
@@ -101,6 +102,19 @@ Currently everything should be compatible with **PHP 5.3-8.0**. _5.2 support is 
 
 They may also contain some fixes, either for long-standing bugs, which haven't made their way into zf1 official repo before EOL, or newly found ones
 and (backwards compatible) adjustments (optimisations for composer autoloader mostly). Maybe even one or two new features.
+
+Still, the main purpose is to allow working on legacy projects on more modern systems, while opening the possibility to **migrate away from zf1 gradually, one component at a time**.
+
+
+### Alternative Installation Method
+
+You may also install the whole framework at once, using composer:
+```
+composer require zf1s/zf1
+```
+to easy up the transition period, but please keep in mind the recommended approach is to
+identify and install only the packages you need.
+
 
 ### Changelog: [here](CHANGELOG.md)
 Original README: [click](README.orig.md)
--- a/composer.json
+++ b/composer.json
@@ -117,6 +117,7 @@
         }
     },
     "replace": {
+        "zendframework/zendframework1": ">=1.12.20",
         "zf1s/zend-acl": "self.version",
         "zf1s/zend-amf": "self.version",
         "zf1s/zend-application": "self.version",
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "zf1s/zf1",
-    "description": "Monorepo for zf1s (Zend Framework 1) packages",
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -102,6 +103,19 @@ Currently everything should be compatible with **PHP 5.3-8.0**. _5.2 support is
 They may also contain some fixes, either for long-standing bugs, which haven't made their way into zf1 official repo before EOL, or newly found ones
 and (backwards compatible) adjustments (optimisations for composer autoloader mostly). Maybe even one or two new features.
 
+Still, the main purpose is to allow working on legacy projects on more modern systems, while opening the possibility to **migrate away from zf1 gradually, one component at a time**.
```
partikus, line 106: @falkenhawk what do you think about adding `suggest` section including all `zf1s/*` packages?

composer.json
```diff
@@ -117,6 +117,7 @@
         }
     },
     "replace": {
+        "zendframework/zendframework1": ">=1.12.20",
```
partikus, line 120: why `>=1.12.20`? Maybe we should replace only versions greater than `>1.12.20` 
https://github.com/zendframework/zf1/tags

partikus (changes requested):

composer.json
falkenhawk: it's fine to cover last officially released version. zf1-future and diablomedia/zf1 have both configured it that way too.

README.md
falkenhawk: that might be an overkill 😅

partikus: you're right, let's ignore it

partikus (approved):
Let's do that! I would verify the usage statistics within next few months.

glensc:
it's still not a drop-in to replacement to `zendframework/zendframework1` i think, as paths are different, and require_once calls are gone, and loader is different (but actually works?).

does this solution actually satisfy what @fredericgboutin-yapla was talking about?

glensc:
maybe for zf1-extras to work need to restore "include-path" directives in the zf1s/zf1 package:
- https://getcomposer.org/doc/04-schema.md#include-path

i.e add every package there:
- packages/zend-*/library/Zend

falkenhawk:
> it's still not a drop-in to replacement to `zendframework/zendframework1` i think, as paths are different, and require_once calls are gone, and loader is different (but actually works?).

It should be, as long as it is installed with `composer`. _(manual installation method is not supported because of the packages folder structure, one would have to invent their autoloader rules, for ancient projects where composer is not available, but they may have better luck with e.g. [zf1-future](https://github.com/Shardj/zf1-future/) or https://github.com/magento/zf1/ which keep the original folder structure)_

It should be also possible now to quickly switch between `zf1-future`/`diablomedia/zf1`/... and this `zf1s/zf1` as `replace` rules are the same.

falkenhawk:
As for `zf1-extras`, we cannot support that out of the box. That would have to be forked and adjusted, or `include-path` rules added locally in a project using it.

https://github.com/fredericgboutin-yapla if you manage to couple zf1-extras or other packages with zf1s, please let us know here or in #128

falkenhawk:
As for `zf1-extras`, we cannot support that out of the box. That would have to be forked and adjusted, or `include-path` rules added locally in a project using it.

@fredericgboutin-yapla if you manage to couple zf1-extras or other packages with zf1s, please let us know here or in #128 🙏

glensc:
@falkenhawk I think we can support here zf1-extras without modification, if we add include-path for all projects. the include path directives you add only to this root project, so the split-out packages will not be affected.

PR #134, opened 2022-10-05
Fix: Add monorepo-builder.yml and phpunit.xml.dist to export-ignore

glensc:
This prevents these files from being copied to `zf1s/zf1` install.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+/.git* export-ignore
 /monorepo-builder.yml export-ignore
 /tests export-ignore
 /phpunit.xml.dist export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
+/monorepo-builder.yml export-ignore
 /tests export-ignore
+/phpunit.xml.dist export-ignore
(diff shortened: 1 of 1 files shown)


falkenhawk:
Thanks for being proactive and paying attention to details :)
I suppose even more files/folders could be omitted:
`.github`
`.gitignore`

image

glensc:
indeed, I didn't look `ls -a`. will check again.

and we could go even further and apply similar rules for `packages/*`

glensc:
added your suggestion, and subpackages i think are fine:

```
➔ find -type f|grep -Ev '(\.php|composer.json|README.md|LICENSE.txt)$|Zend/Locale/Data|Zend/Queue/Adapter/Db/|resources/PhpOnAzure.Web/resources/WebPICmdLine/|Zend/Text/Figlet/zend-framework.flf|Scaffolders/DefaultScaffolder.phar|Scaffolders/DefaultScaffolder/resources|./zend-tool/bin/zf.sh|./zend-service-windowsazure/library/Zend/Service/WindowsAzure/CommandLine/Scaffolders/DefaultScaffolder/build.bat'
```

falkenhawk (approved):
perfect, thanks!

PR #138, opened 2022-10-11
Reduce redundant else-branching (after throws), add missing @throws annotations

glensc:
Carry https://github.com/Shardj/zf1-future/pull/104.

- Excluded whitespace and formatting changes.
- Included extra fixes:
   - https://github.com/Shardj/zf1-future/pull/104/files#r992515043
   - https://github.com/Shardj/zf1-future/pull/104/files#r992486395
   
cc @kstenschke

### DIFF
--- a/packages/zend-amf/library/Zend/Amf/Server.php
+++ b/packages/zend-amf/library/Zend/Amf/Server.php
@@ -212,6 +212,7 @@ class Zend_Amf_Server implements Zend_Server_Interface
     /**
      * @param namespace of all incoming sessions defaults to Zend_Amf
      * @return Zend_Amf_Server
+     * @throws Zend_Session_Exception
      */
     public function setSession($namespace = 'Zend_Amf')
     {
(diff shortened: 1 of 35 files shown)


glensc:
php 8.0 job needs retry. i don't have permission to do so myself. some temporary error setting up container.

falkenhawk:
@glensc `@throws` tags are useful indeed.

But do we really need all the code beautifying? let the zf1 be the ugly beast, only taming it to make it work without errors on newer php versions. I don't want to follow others path trying to make the codebase look cleaner or more modern :)

falkenhawk:
or is it a matter of a IDE complaining there, for redundant `else` after `return`, when using zf1 in a project?

glensc:
I already excluded whitespace and formatting changes (if you open the original pull request), which makes merging the codebase more difficult again, but at least `diff -w` might help (haven't tested)

glensc:
Extracted annotations:
- https://github.com/zf1s/zf1/pull/140

so you can decide about the other changes later.

falkenhawk:
Oh that's lovely, thanks @glensc . I will look into it

falkenhawk:
decided to go with #140 instead

PR #139, opened 2022-10-11
Fix CI warnings: Node.js 12 actions are deprecated.

glensc:
For more information see:
- https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

Please update the following actions to use Node.js 16: actions/checkout, actions/checkout


falkenhawk (approved):
👏

PR #142, opened 2022-10-14
fix: github actions' failures

partikus:


falkenhawk (approved):
looks nice, easier to understand, thanks!

PR #141, opened 2022-10-14
RE: Added \ReturnTypeWillChange Attribute - (#120)

partikus:
This PR updates #120 and switches to `zf1s/polyfill-php81`.

All credits goes to @jack-worman 🥇

PR #143, opened 2022-11-15
Fix str_split('') logic to keep same as PHP before 8.2 on PHP 8.2

glensc:
_Carries  https://github.com/Shardj/zf1-future/pull/277 which picked only relevant change from https://github.com/Shardj/zf1-future/pull/275 which was created by @hungtrinh_

Reference:
-  https://php.watch/versions/8.2/str_split-empty-string-empty-array

Fixes:
- https://github.com/Shardj/zf1-future/issues/276


falkenhawk:
Thanks! We will look into it after #141 and #145

glensc:
Unrelated to https://github.com/zf1s/zf1/pull/141 and https://github.com/zf1s/zf1/pull/145 both

partikus (approved):

partikus:
@falkenhawk IMHO looks fine

https://3v4l.org/mHMvb

PR #144, opened 2022-11-15
Add compat symlinks for original zendframework layout

glensc:
This allows easier patching and syncing with [zf1 future](https://github.com/Shardj/zf1-future):

In `zf1-future` repo:
1. Extract patch from the other repo: `git format-patch  -1 COMMIT`

In `zf1s` repo:
1. switch to this branch `git checkout symlink-compat`
2. create new branch: `git checkout -b zf1-future/asdf`
3. and apply the patch file `git am 0001-something.patch`
4. rebase without the symlinks commit `git rebase -i origin/master`

an alternative would be to use `cherry-pick`:
1. `git remote add zf1-future https://github.com/Shardj/zf1-future`
2. `git fetch zf1-future`
3. `git cherry-pick COMMIT`

### DIFF
--- a/library/Zend/Acl
+++ b/library/Zend/Acl
@@ -1,0 +1,1 @@
+../../packages/zend-acl/library/Zend/Acl
--- a/library/Zend/Acl.php
+++ b/library/Zend/Acl.php
@@ -1,0 +1,1 @@
+../../packages/zend-acl/library/Zend/Acl.php
--- a/library/Zend/Amf
+++ b/library/Zend/Amf
@@ -1,0 +1,1 @@
+../../packages/zend-amf/library/Zend/Amf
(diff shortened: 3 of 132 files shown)


partikus:
@glensc did you consider to prepare a php script for creating symlinks of legacy layout? e.g. `composer create-symlinks` sounds interesting and then we could keep it in the repo

glensc:
@partikus no I did not consider it:
- https://github.com/zf1s/zf1/pull/144#issuecomment-1315711855

are the symlinks even necessary?

why do you want the symlinks? if there's a use case, perhaps it's just worth merging.

partikus:
TBH I do not need them at all ;) I've written that question with losing context about `Shardj/zf1-future`. Did you update description or I've overlooked it? Does not matter. Anyway, thank you for responding.

PR #145, opened 2022-11-15
Add AllowDynamicProperties Attribute to classes

glensc:
_Carry https://github.com/Shardj/zf1-future/pull/278, which was extracted from https://github.com/Shardj/zf1-future/pull/275 which was created by @hungtrinh_:

> To save time, i used [Classes with #[AllowDynamicProperties] attribute](https://php.watch/versions/8.2/dynamic-properties-deprecated#AllowDynamicProperties) to resolve issue Deprecated: Creation of dynamic property {x} is deprecated on PHP 8.2

This is to fix the issue:
> `Deprecated: Creation of dynamic property {x} is deprecated`

### DIFF
--- a/packages/zend-application/library/Zend/Application/Bootstrap/Bootstrap.php
+++ b/packages/zend-application/library/Zend/Application/Bootstrap/Bootstrap.php
@@ -37,6 +37,7 @@
  * @copyright  Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
  * @license    http://framework.zend.com/license/new-bsd     New BSD License
  */
+#[AllowDynamicProperties]
 class Zend_Application_Bootstrap_Bootstrap
     extends Zend_Application_Bootstrap_BootstrapAbstract
 {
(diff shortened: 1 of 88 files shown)


falkenhawk:
Thank you for setting this PR! This is going to greatly reduce the amount of deprecation warnings in 8.2. We will tackle this after #141 is finalized, hopefully within next two weeks.

falkenhawk (changes requested):
On second thought I'd prefer adding missing props, instead of taking a shortcut with adding only class attributes, as this would be a temporary workaround which would throw fatal errors in future php versions.

I took the approach with adding missing attributes in https://github.com/ovos/doctrine1/pull/12 and I've found a few inconsistencies along the way, so I consider those deprecation warnings as one of the most useful ones.

hungtrinh:
I also want to add the missing props that's why I created https://github.com/Shardj/zf1-future/pull/291 to fix my mess

glensc:
Needs to include https://github.com/Shardj/zf1-future/pull/291 as well, back to Draft

glensc:
This is incomplete. should start over porting from z1-future

PR #146, opened 2022-11-18
Add zf1s/polyfill-php81 dependencies

glensc:
_Extracted from https://github.com/zf1s/zf1/pull/141_

- [x] removed composer sort option
- [x] sorted "require" list
- [x] updated "require" only where needed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,8 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=5.3.3",
-        "symfony/polyfill-php70": "^1.19"
+        "symfony/polyfill-php70": "^1.19",
+        "zf1s/polyfill-php81": "^1.25"
     },
     "require-dev": {
         "ext-ctype": "*",
--- a/packages/zend-acl/composer.json
+++ b/packages/zend-acl/composer.json
@@ -5,6 +5,7 @@
     "homepage": "http://framework.zend.com/",
     "license": "BSD-3-Clause",
     "require": {
+        "zf1s/polyfill-php81": "^1.25",
         "php": ">=5.3.3",
         "zf1s/zend-exception": "^1.15.0"
     },
--- a/packages/zend-acl/composer.json
+++ b/packages/zend-acl/composer.json
@@ -5,7 +5,6 @@
     "homepage": "http://framework.zend.com/",
     "license": "BSD-3-Clause",
     "require": {
-        "zf1s/polyfill-php81": "^1.25",
         "php": ">=5.3.3",
         "zf1s/zend-exception": "^1.15.0"
     },
--- a/packages/zend-amf/composer.json
+++ b/packages/zend-amf/composer.json
@@ -5,7 +5,6 @@
     "homepage": "http://framework.zend.com/",
     "license": "BSD-3-Clause",
     "require": {
-        "zf1s/polyfill-php81": "^1.25",
         "php": ">=5.3.3",
         "ext-dom": "*",
         "zf1s/zend-exception": "^1.15.0",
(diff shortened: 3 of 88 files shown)


glensc:
> but could you rearrange requires in individual package so that they are sorted alphabetically, please?
> And please make sure polyfill-php81 is required only when it is really needed for a package, i.e. when any of the following are used inside the package.
> 
> [array_is_list](https://php.net/array_is_list)
> [MYSQLI_REFRESH_REPLICA](https://www.php.net/manual/en/mysqli.constants.php#constantmysqli-refresh-replica) constant
> [ReturnTypeWillChange](https://wiki.php.net/rfc/internal_method_return_types)
>
> But after a short research, it might be that we don't need the polyfill at all? First two features are not used anywhere, if I'm not mistaken, and ReturnTypeWillChange stub would be only needed for php 8.0, IF it would throw an error on it. But it does not throw an error, either for that or any attribute which is not implemented, e.g. #[Gibberish] would not throw any error even if class Gibberish is not implemented as an attribute.
> 
> Maybe it would be needed for code analysis, running on php 8.0. Even in that case, it would be probably enough, if added to the main composer.json only. But we do not have any code analyzers running here, and if one dev would like to run it locally, why would they run it on php 8.0?
> 
> Please check if those polyfills are really needed. thank you

as you said, array_is_list MYSQLI_REFRESH_REPLICA not present, but ReturnTypeWillChange from https://github.com/zf1s/zf1/pull/141 are:

```
git grep -l ReturnTypeWillChange|cut -d/ -f1|sort -u|sed -e 's/^/- [x] /'
```

- [x] zend-cloud
- [x] zend-config
- [x] zend-controller
- [x] zend-crypt
- [x] zend-db
- [x] zend-dojo
- [x] zend-dom
- [x] zend-eventmanager
- [x] zend-feed
- [x] zend-file
- [x] zend-form
- [x] zend-gdata
- [x] zend-http
- [x] zend-ldap
- [x] zend-mail
- [x] zend-markup
- [x] zend-memory
- [x] zend-navigation
- [x] zend-paginator
- [x] zend-pdf
- [x] zend-progressbar
- [x] zend-queue
- [x] zend-reflection
- [x] zend-rest
- [x] zend-search-lucene
- [x] zend-server
- [x] zend-service-amazon
- [x] zend-service-delicious
- [x] zend-service-ebay
- [x] zend-service-flickr
- [x] zend-service-rackspace
- [x] zend-service-yahoo
- [x] zend-session
- [x] zend-stdlib
- [x] zend-tag
- [x] zend-timesync
- [x] zend-tool
- [x] zend-uri
- [x] zend-view

glensc:
@falkenhawk @partikus thanks for the input. i read way diagonally https://github.com/zf1s/zf1/pull/141 notes.

made local test:

```
zf1s (feature/php-8-1-compat) ➜ cat ReturnTypeWillChange-test.php
<?php

class a {
     #[\ReturnTypeWillChange]
     public function getIterator() {}
}

zf1s (feature/php-8-1-compat) ➜ php --version
PHP 8.0.25 (cli) (built: Oct 31 2022 08:58:37) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.25, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.25, Copyright (c), by Zend Technologies

zf1s (feature/php-8-1-compat) ➜ php -l ReturnTypeWillChange-test.php
No syntax errors detected in ReturnTypeWillChange-test.php

zf1s (feature/php-8-1-compat) ➜ php ReturnTypeWillChange-test.php
zf1s (feature/php-8-1-compat) ➜
```

this pr is not needed at all, if `zf1s/polyfill-php81` is not added to deps, there's nothing to sort either

PR #149, opened 2022-11-28
php 8.1 compatibility fixes

falkenhawk:
- fix deprecation notices for `Serializable` interface
  `Deprecated: Test implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary)`
  https://php.watch/versions/8.1/serializable-deprecated
- [zend-amf] fix for php 8.1+ keep the order of properties when they are being serialized same as PHP <8.1

  In PHP 8.1, the order of properties is the same as in the class definition, but properties from parent class always go after properties from the current class.
  From PHP upgrade notes: https://github.com/php/php-src/blob/578b67da49af51b2f796a48782e51ceb62860943/UPGRADING#L334-L341

  Copied the `correlationId` property definition from `Zend_Amf_Value_Messaging_AsyncMessage` to parent `Zend_Amf_Value_Messaging_AbstractMessage`.
  `correlationId` prop is expected to be at the beginning of serialized messages, otherwise it breaks Zend_Amf_ResponseTest cases for php 8.1+.
  
  This was causing 14 failures in `Zend_Amf` test suite due to serialized binary data was different than expected, due to changed order of properties of serialized objects.

- [zend-server] fix issue with `Zend_Server_Reflection_Method`

  Method name was not set in constructor, but it was being read in tests (there is no property defined for it, but `__set` magic method is defined in the base class)
  resulting in `substr(): Passing null to parameter https://github.com/zf1s/zf1/pull/1 ($string) of type string is deprecated` and similar errors in various places

- [fixing errors and warnings for php 8.1](https://github.com/zf1s/zf1/pull/149/commits/37d353f8e6af3c629fdea7828ad48ef1f3c71f4b)

- includes #151 and #156

### DIFF
--- a/packages/zend-db/library/Zend/Db/Statement/Pdo.php
+++ b/packages/zend-db/library/Zend/Db/Statement/Pdo.php
@@ -111,12 +111,7 @@ class Zend_Db_Statement_Pdo extends Zend_Db_Statement implements IteratorAggrega
                     $type = PDO::PARAM_STR;
                 }
             }
-            if ($length === null) {
-                // PDOStatement::bindParam(): Passing null to parameter #4 ($maxLength) of type int is deprecated
-                // since php 8.1.0
-                $length = 0;
-            }
-            return $this->_stmt->bindParam($parameter, $variable, $type, $length, $options);
+            return $this->_stmt->bindParam($parameter, $variable, $type, (int) $length, $options);
         } catch (PDOException $e) {
             // require_once 'Zend/Db/Statement/Exception.php';
             throw new Zend_Db_Statement_Exception($e->getMessage(), $e->getCode(), $e);
--- a/packages/zend-db/library/Zend/Db/Statement/Pdo.php
+++ b/packages/zend-db/library/Zend/Db/Statement/Pdo.php
@@ -111,6 +111,11 @@ class Zend_Db_Statement_Pdo extends Zend_Db_Statement implements IteratorAggrega
                     $type = PDO::PARAM_STR;
                 }
             }
+            if ($length === null) {
+                // PDOStatement::bindParam(): Passing null to parameter #4 ($maxLength) of type int is deprecated
+                // since php 8.1.0
+                $length = 0;
+            }
             return $this->_stmt->bindParam($parameter, $variable, $type, $length, $options);
         } catch (PDOException $e) {
             // require_once 'Zend/Db/Statement/Exception.php';
@@ -245,10 +250,13 @@ class Zend_Db_Statement_Pdo extends Zend_Db_Statement implements IteratorAggrega
      * @return mixed Array, object, or scalar depending on fetch mode.
      * @throws Zend_Db_Statement_Exception
      */
-    public function fetch($style = null, $cursor = null, $offset = null)
+    public function fetch($style = null, $cursor = null, $offset = 0)
     {
         if ($style === null) {
             $style = $this->_fetchMode;
+        }
+        if ($cursor === null) {
+            $cursor = PDO::FETCH_ORI_NEXT;
         }
         try {
             return $this->_stmt->fetch($style, $cursor, $offset);
--- a/packages/zend-db/library/Zend/Db/Statement/Pdo/Ibm.php
+++ b/packages/zend-db/library/Zend/Db/Statement/Pdo/Ibm.php
@@ -83,7 +83,7 @@ class Zend_Db_Statement_Pdo_Ibm extends Zend_Db_Statement_Pdo
             if (($type === null) && ($length === null) && ($options === null)) {
                 return $this->_stmt->bindParam($parameter, $variable);
             } else {
-                return $this->_stmt->bindParam($parameter, $variable, $type, $length === null ? 0 : $length, $options);
+                return $this->_stmt->bindParam($parameter, $variable, $type, (int) $length, $options);
             }
         } catch (PDOException $e) {
             // require_once 'Zend/Db/Statement/Exception.php';
--- a/packages/zend-db/library/Zend/Db/Statement/Pdo/Ibm.php
+++ b/packages/zend-db/library/Zend/Db/Statement/Pdo/Ibm.php
@@ -83,7 +83,7 @@ class Zend_Db_Statement_Pdo_Ibm extends Zend_Db_Statement_Pdo
             if (($type === null) && ($length === null) && ($options === null)) {
                 return $this->_stmt->bindParam($parameter, $variable);
             } else {
-                return $this->_stmt->bindParam($parameter, $variable, $type, $length, $options);
+                return $this->_stmt->bindParam($parameter, $variable, $type, $length === null ? 0 : $length, $options);
             }
         } catch (PDOException $e) {
             // require_once 'Zend/Db/Statement/Exception.php';
--- a/packages/zend-file-transfer/library/Zend/File/Transfer/Adapter/Abstract.php
+++ b/packages/zend-file-transfer/library/Zend/File/Transfer/Adapter/Abstract.php
@@ -185,7 +185,7 @@ abstract class Zend_File_Transfer_Adapter_Abstract
      */
     public function setPluginLoader(Zend_Loader_PluginLoader_Interface $loader, $type)
     {
-        $type = is_string($type) ? strtoupper($type) : $type;
+        $type = strtoupper((string) $type);
         switch ($type) {
             case self::FILTER:
             case self::VALIDATE:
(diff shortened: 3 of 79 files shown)


falkenhawk:
236 errors, 9 failures to go... 😵‍💫

falkenhawk:
only 6 errors left. all in `zend-db`

falkenhawk:
@glensc would you like to review this PR?

packages/zend-db/library/Zend/Db/Statement/Pdo/Ibm.php
```diff
@@ -83,7 +83,7 @@ public function _bindParam($parameter, &$variable, $type = null, $length = null,
             if (($type === null) && ($length === null) && ($options === null)) {
                 return $this->_stmt->bindParam($parameter, $variable);
             } else {
-                return $this->_stmt->bindParam($parameter, $variable, $type, $length, $options);
+                return $this->_stmt->bindParam($parameter, $variable, $type, $length === null ? 0 : $length, $options);
```
partikus, line 86: why we cannot just use `(int) $length`, can we?

packages/zend-db/library/Zend/Db/Statement/Pdo.php
```diff
@@ -111,6 +111,11 @@ protected function _bindParam($parameter, &$variable, $type = null, $length = nu
                     $type = PDO::PARAM_STR;
                 }
             }
+            if ($length === null) {
+                // PDOStatement::bindParam(): Passing null to parameter #4 ($maxLength) of type int is deprecated
+                // since php 8.1.0
+                $length = 0;
```
partikus, line 117: `(int) $length` ?  👀

packages/zend-file-transfer/library/Zend/File/Transfer/Adapter/Abstract.php
```diff
@@ -185,7 +185,7 @@ public static function getProgress()
      */
     public function setPluginLoader(Zend_Loader_PluginLoader_Interface $loader, $type)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 188: `$type = strtoupper((string) $type);` ?

packages/zend-file-transfer/library/Zend/File/Transfer/Adapter/Abstract.php
```diff
@@ -209,7 +209,7 @@ public function setPluginLoader(Zend_Loader_PluginLoader_Interface $loader, $typ
      */
     public function getPluginLoader($type)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 212: idem

packages/zend-file-transfer/library/Zend/File/Transfer/Adapter/Abstract.php
```diff
@@ -256,7 +256,7 @@ public function getPluginLoader($type)
      */
     public function addPrefixPath($prefix, $path, $type = null)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 259: idem

packages/zend-form/library/Zend/Form.php
```diff
@@ -417,7 +417,7 @@ public function setConfig(Zend_Config $config)
      */
     public function setPluginLoader(Zend_Loader_PluginLoader_Interface $loader, $type = null)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 420: idem

packages/zend-form/library/Zend/Form.php
```diff
@@ -445,7 +445,7 @@ public function setPluginLoader(Zend_Loader_PluginLoader_Interface $loader, $typ
      */
     public function getPluginLoader($type = null)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 448: idem

packages/zend-form/library/Zend/Form.php
```diff
@@ -495,7 +495,7 @@ public function getPluginLoader($type = null)
      */
     public function addPrefixPath($prefix, $path, $type = null)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 498: idem

packages/zend-form/library/Zend/Form/Decorator/Description.php
```diff
@@ -159,7 +159,7 @@ public function render($content)
         }
 
         $description = $element->getDescription();
-        $description = trim($description);
+        $description = is_string($description) ? trim($description) : '';
```
partikus, line 162: `$description = trim((string) $description);`

packages/zend-form/library/Zend/Form/Element/Captcha.php
```diff
@@ -201,7 +201,7 @@ public function render(Zend_View_Interface $view = null)
      */
     public function getPluginLoader($type)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 204: idem

packages/zend-form/library/Zend/Form/Element/Captcha.php
```diff
@@ -228,7 +228,7 @@ public function getPluginLoader($type)
      */
     public function addPrefixPath($prefix, $path, $type = null)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 231: idem

packages/zend-form/library/Zend/Form/Element/File.php
```diff
@@ -104,7 +104,7 @@ public function loadDefaultDecorators()
      */
     public function setPluginLoader(Zend_Loader_PluginLoader_Interface $loader, $type)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 107: idem

packages/zend-form/library/Zend/Form/Element/File.php
```diff
@@ -122,7 +122,7 @@ public function setPluginLoader(Zend_Loader_PluginLoader_Interface $loader, $typ
      */
     public function getPluginLoader($type)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 125: idem

packages/zend-form/library/Zend/Form/Element/File.php
```diff
@@ -149,7 +149,7 @@ public function getPluginLoader($type)
      */
     public function addPrefixPath($prefix, $path, $type = null)
     {
-        $type = strtoupper($type);
+        $type = is_string($type) ? strtoupper($type) : $type;
```
partikus, line 152: idem

partikus (changes requested):
@falkenhawk looks really good, few minor CRs added

packages/zend-file-transfer/library/Zend/File/Transfer/Adapter/Abstract.php
partikus: oh, for `addPrefixPath` we have to allow passing `null`, so it's not relevant anymore

packages/zend-form/library/Zend/Form.php
partikus: oh, for `addPrefixPath` we have to allow passing `null`, so it's not relevant anymore

packages/zend-form/library/Zend/Form/Element/Captcha.php
partikus: oh, for `addPrefixPath` we have to allow passing `null`, so it's not relevant anymore

packages/zend-form/library/Zend/Form/Element/Captcha.php
falkenhawk: I will leave this one as it calls `parent::getPluginLoader($type)` and theoretically that could be called with `null` and handled there.

packages/zend-form/library/Zend/Form/Element/File.php
falkenhawk: I will leave this one as it calls `parent::setPluginLoader($type)` and potentially that could be called with null and handled there.

packages/zend-form/library/Zend/Form/Element/File.php
falkenhawk: I will leave this one as it calls `parent::getPluginLoader($type)` and potentially that could be called with `null` and handled there.

partikus: ok, thank you for taking care about BC 👍

partikus (approved):

marcing (approved):

PR #151, opened 2022-11-29
[zend-date] properly calculate sunrise, sunset and twilight times

falkenhawk:
`date_sunset()` and `date_sunrise()` are deprecated in php 8.1+ - switched to recommended `date_sun_info()` function (which is available since php v5.1)
https://www.php.net/manual/en/function.date-sun-info.php#refsect1-function.date-sun-info-returnvalues

but that function yields slightly different results since zenith angles for sunrise/sunset/twilights are hardcoded internally
and they are different to what zf used before (see `$horizonDeclination` in `Zend_Date::calcSun()` - moved from `Zend_Date::_checkLocation()`)

Yet ONLY NOW they are accurate! (calculations for civil / nautical / astronomical twilight were totally wrong before)

### DIFF
--- a/tests/Zend/Date/DateObjectTest.php
+++ b/tests/Zend/Date/DateObjectTest.php
@@ -284,77 +284,120 @@ class Zend_Date_DateObjectTest extends PHPUnit_Framework_TestCase
         // (which was later reverted in php < 7.2.0)
         // Example of the difference: https://3v4l.org/v46rk
         // Not really something we can test the same in all versions, so doing a version_compare here.
-        if (PHP_VERSION_ID >= 70200) {
-            $this->assertSame( 9961716, $date->calcSun(array('latitude' =>  38.4, 'longitude' => -29), -0.0145439, true ));
-            $this->assertSame(10010341, $date->calcSun(array('latitude' =>  38.4, 'longitude' => -29), -0.0145439, false));
-            $this->assertSame( 9966981, $date->calcSun(array('latitude' => -38.4, 'longitude' => -29), -0.0145439, true ));
-            $this->assertSame(10005077, $date->calcSun(array('latitude' => -38.4, 'longitude' => -29), -0.0145439, false));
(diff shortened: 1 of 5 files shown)


marcing (approved):
Thank you for being so thorough with this fix.

tests/Zend/Date/DateObjectTest.php
```diff
@@ -284,77 +284,120 @@ public function testCalcSunInternal()
...
-            $this->assertSame(-148293174, $date->calcSun(array('latitude' => -38.4, 'longitude' =>  29), -0.0145439, false));
-            $this->assertSame(-148360510, $date->calcSun(array('latitude' =>  38.4, 'longitude' => 129), -0.0145439, true ));
-            $this->assertSame(-148312728, $date->calcSun(array('latitude' =>  38.4, 'longitude' => 129), -0.0145439, false));
-            $this->assertSame(-148356087, $date->calcSun(array('latitude' => -38.4, 'longitude' => 129), -0.0145439, true ));
-            $this->assertSame(-148317151, $date->calcSun(array('latitude' => -38.4, 'longitude' => 129), -0.0145439, false));
-            $this->assertSame(-148298649, $date->calcSun(array('latitude' =>  38.4, 'longitude' =>-129), -0.0145439, true ));
-            $this->assertSame(-148250768, $date->calcSun(array('latitude' =>  38.4, 'longitude' =>-129), -0.0145439, false));
-            $this->assertSame(-148294127, $date->calcSun(array('latitude' => -38.4, 'longitude' =>-129), -0.0145439, true ));
-            $this->assertSame(-148255290, $date->calcSun(array('latitude' => -38.4, 'longitude' =>-129), -0.0145439, false));
+        if (PHP_VERSION_ID >= 80100) {
```
falkenhawk, line 350:
```suggestion
        if (PHP_VERSION_ID >= 80000) {
```

PR #153, opened 2022-12-02
[zend-timesync] fix ntp time sync

falkenhawk:
NTP timestamps are represented as a 64-bit fixed-point number, in seconds relative to 0000 UT on 1 January 1900.  The integer part is in the first 32 bits and the fraction part in the last 32 bits.

https://www.eecis.udel.edu/~mills/time.html
https://tickelton.gitlab.io/articles/ntp-timestamps/

But calculation of the fractional part was broken since forever, always sending 00000000 to a ntp server - fractional seconds were not being multiplied by max value of 32-bit+1 to represent the fraction in 0-[(2^32)-1] range, as expected by ntp protocol. In result, Zent_TimeSync_Ntp has always been reporting time in full-second resolution to ntp servers when syncing time.

Moreover, it triggered `Implicit conversion from float-string "0.xyz" to int loses precision` in `Zend/TimeSync/Ntp.php:70` on PHP 8.1 and that's how I started my investigation.

### DIFF
--- a/packages/zend-timesync/library/Zend/TimeSync/Ntp.php
+++ b/packages/zend-timesync/library/Zend/TimeSync/Ntp.php
@@ -56,6 +56,37 @@ class Zend_TimeSync_Ntp extends Zend_TimeSync_Protocol
     }
 
     /**
+     * Convert microtime (result of microtime() as string "msec sec") to NTP timestamp
+     *
+     * NTP timestamps are represented as a 64-bit fixed-point number, in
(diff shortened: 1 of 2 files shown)


falkenhawk:
for the record, I compared the logic with java implementation [TimeStamp::toNtpTime()](https://commons.apache.org/proper/commons-net/javadocs/api-1.4.1/src-html/org/apache/commons/net/ntp/TimeStamp.html)

PR #155, opened 2022-12-05
fix "stty: 'standard input': Inappropriate ioctl for device" spam

falkenhawk:
image


partikus (approved):
An interesting approach with redirecting `stderr` to `/dev/null`. 👍

falkenhawk:
@marcinlee suggested that `[[ $- == *i* ]] && stty size` could be used instead - to avoid calling `stty` at all when shell is not interactive. Might be cleaner than to discarding stderr. 🤔

partikus:
@falkenhawk both are fine, IMHO discarding `stderr` is easier to understand though the condition looks more appropriate approach

PR #156, opened 2022-12-06
[zend-db] php 8.1 - tests adjustments 

partikus:
https://php.watch/versions/8.1/mysqli-error-mode


packages/zend-db/library/Zend/Db/Statement/Mysqli.php
```diff
@@ -67,7 +67,13 @@ public function _prepare($sql)
     {
         $mysqli = $this->_adapter->getConnection();
 
-        $this->_stmt = $mysqli->prepare($sql);
+        // https://www.php.net/manual/en/class.mysqli-sql-exception.php
+        // https://php.watch/versions/8.1/mysqli-error-mode
+        try {
+            $this->_stmt = $mysqli->prepare($sql);
+        } catch (\mysqli_sql_exception $e) {
+            throw \Zend_Db_Statement_Mysqli_Exception::fromMysqliException($e);
```
falkenhawk, line 75: There is no need to use namespace qualifier as zf1 is not namespaced.
```suggestion
        } catch (mysqli_sql_exception $e) {
            throw Zend_Db_Statement_Mysqli_Exception::fromMysqliException($e);
```

packages/zend-db/library/Zend/Db/Statement/Mysqli.php
```diff
@@ -67,7 +67,13 @@ public function _prepare($sql)
     {
         $mysqli = $this->_adapter->getConnection();
 
-        $this->_stmt = $mysqli->prepare($sql);
+        // https://www.php.net/manual/en/class.mysqli-sql-exception.php
+        // https://php.watch/versions/8.1/mysqli-error-mode
```
falkenhawk, line 71: I think these comments with links here are not necessary. It will be difficult to guess the context and why they are here for someone who takes a look in that part of the code in the future.
Handling exception is valid for any php version, not only 8.1+.
In my opinion it is enough to mention it in the PR description. 
Otherwise it would need additional context here and be copied to every occurrence in the code where mysqli exceptions are being handled now.

partikus:
@falkenhawk updated, thank you for your comments 👍

packages/zend-db/library/Zend/Db/Adapter/Mysqli.php
```diff
@@ -343,6 +343,14 @@ protected function _connect()
...
             mysqli_set_charset($this->_connection, $this->_config['charset']);
         }
+
+        if (!isset($this->_config['report_mode']) || PHP_VERSION_ID >= 80100) {
+            mysqli_report(
+                isset($this->_config['report_mode'])
+                    ? $this->_config['report_mode']
+                    : MYSQLI_REPORT_OFF
+            );
+        }
```
falkenhawk, line 353: I suppose this should not be added since exceptions are being handled now?

partikus: right 👍  fixed

packages/zend-db/library/Zend/Db/Adapter/Mysqli/Exception.php
```diff
@@ -37,4 +37,16 @@
  */
 class Zend_Db_Adapter_Mysqli_Exception extends Zend_Db_Adapter_Exception
 {
+    public static function fromMysqliException($exception)
```
glensc, line 40: typing could be added: `public static function fromMysqliException(mysqli_sql_exception $exception)`

PR #157, opened 2022-12-06
[github-actions] enable SQLite tests

partikus:
since `pdo_sqlite` extension is installed, we could start supporting `sqlite` tests on CI

Since [PHP 8.1](https://www.php.net/manual/en/migration81.incompatible.php#migration81.incompatible.pdo.sqlite
) introduced BC break while returning php types instead of strings, we have to revert the previous configuration.
Overwritten attribute `\PDO::ATTR_STRINGIFY_FETCHES => true` can be still changed through the `['driver_options' => [\PDO::ATTR_STRINGIFY_FETCHES => false]]` config for sqlite adapter.

### DIFF
--- a/packages/zend-db/library/Zend/Db/Adapter/Pdo/Sqlite.php
+++ b/packages/zend-db/library/Zend/Db/Adapter/Pdo/Sqlite.php
@@ -91,6 +91,14 @@ class Zend_Db_Adapter_Pdo_Sqlite extends Zend_Db_Adapter_Pdo_Abstract
         $this->_config['username'] = null;
         $this->_config['password'] = null;
 
+        if (PHP_VERSION_ID >= 80100) {
+            // ensure $config['driver_options'] is an array
+            $config['driver_options'] = empty($config['driver_options']) ? array() : $config['driver_options'];
+            if (!isset($config['driver_options'][PDO::ATTR_STRINGIFY_FETCHES])) {
+                $config['driver_options'][PDO::ATTR_STRINGIFY_FETCHES] = true;
+            }
+        }
+
         return parent::__construct($config);
     }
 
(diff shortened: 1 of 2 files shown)


falkenhawk:
@partikus please rebase this branch onto `php81-compat` branch

packages/zend-db/library/Zend/Db/Adapter/Pdo/Sqlite.php
```diff
@@ -91,6 +91,13 @@ public function __construct(array $config = array())
         $this->_config['username'] = null;
         $this->_config['password'] = null;
 
+        if (PHP_VERSION_ID >= 80100) {
+            $config['driver_options'] = empty($config['driver_options']) ? array() : $config['driver_options']; // ensure $config['driver_options'] is an array
```
glensc, line 95: keep code and comments on separate line (put comment before the code)

falkenhawk:
for the record:
before: `Tests: 16125, Assertions: 63458, Incomplete: 21, Skipped: 591.`
after:  `Tests: 16676, Assertions: 65023, Incomplete: 24, Skipped: 609.`

falkenhawk:
@partikus could we also include https://github.com/Shardj/zf1-future/pull/284 please?

glensc:
Carrying it: https://github.com/zf1s/zf1/pull/160

glensc (approved):

falkenhawk (approved):
I think it's the correct approach to keep the current behavior also on php 8.1 instead of adjusting text assertions (as long as it is feasible) - end users may still adjust their config for zend-db + sqlite if they want to follow the behavior of php 8.1+.
Thank you @partikus for dealing with this.

Issue #165, opened 2023-02-13
Zend Session would throw an exception "The session has already been started. The session id must be set first." upon receiving an invalid session ID


fredericgboutin-yapla:
Long story short,
We monitored a lot of exceptions this morning on our website, all stating "The session has already been started. The session id must be set first."

There are a lot of different reasons why this message can happen.

Our logs indicated that the user ID values were in fact all attempting to break things up, with `PHPSESSID` cookie  having values like:

- `Thread.Sleep(4*1000);`
- `C:/winnt/win.ini`
- `${T(java.lang.System).properties}`

Etc.

Upon analysis I stumbled on this old issue https://github.com/zendframework/zend-session/issues/119 which helped me to reproduce the issue on my dev. env. locally with a simple `curl` command like suggested,

`curl -I 'http://your-local-website.com/' -H 'Cookie: PHPSESSID=_test_'`

I then Xdebug-ed it to realize that Zend doesn't properly account for the situation when `session_id` is given a rejected ID. In that situation, the `SID` constant is defined BUT its value is an empty string (at least, on PHP 7.0.33),

> session_id() returns the session id for the current session or the empty string ("") if there is no current session
>
> Source: https://www.php.net/manual/en/function.session-id.php

fredericgboutin-yapla:
When trying to unit-test my fix, I realized that, as a workaround, one can simply call `session_id($_COOKIE['PHPSESSID'] ?? null);` right BEFORE the first call to `Zend_Session::start()`

As bonus, doing so prevents the creation of sessions with wrong session ID at the first place.

fredericgboutin-yapla:
Since explicitly sending an invalid `PHPSESSID`makes Zend create 2 sessions (but only retaining one), I realize that an attacker could use this flaw to prematurely close active sessions, depending on the type of session storage in used - i .e. one `curl` request in my example momentarily occupy 2 sessions on the storage.

For example, when using a Redis setup or any FIFO session storage.

fredericgboutin-yapla:
Fun fact,
After reviewing PHP's internal code to understand how the `SID` constant was created and why / how its value diverged from `session_id()` I realized that PHP already tries to cleanup "dangerous characters" in the session ID - https://github.com/php/php-src/blob/PHP-7.0.33/ext/session/session.c#L1684 - 

This logic doesn't do very much to help here, unfortunately.

fredericgboutin-yapla:
Another fun fact,
[...]

PR #167, opened 2023-02-13
Fix invalid session regeneration on wrong session ID

fredericgboutin-yapla:
Fix for https://github.com/zf1s/zf1/issues/165

Zend have a logic in place to create a valid session ID from the orignal value, MD5-ing it. This logic doesn't properly kick in the aforementioned situation of the issue but with this fix it does.

The fix does NOT prevent an initial write through the Session handler. This means that session with invalid keys are created into the database, file or redis instance in use for session storage.

In order to fix that, I'm afraid we would need to refactor the code to always md5 invalid entries pro-actively and since `getId()` is empty at that moment, I'm really not sure how we should to that.

Note that the comment `// Check to see if we've been passed an invalid session ID` at the beginning of the `start()` method only really apply on an already opened session.

### DIFF
--- a/packages/zend-session/library/Zend/Session.php
+++ b/packages/zend-session/library/Zend/Session.php
@@ -426,7 +426,7 @@ class Zend_Session extends Zend_Session_Abstract
             // Generate a valid, temporary replacement
             self::setId(md5(self::getId()));
             // Force a regenerate after session is started
-            self::$_regenerateIdState = -1;
+            self::regenerateId();
         }
 
         if (self::$_sessionStarted && self::$_destroyed) {
@@ -459,7 +459,7 @@ class Zend_Session extends Zend_Session_Abstract
         }
 
         // See http://www.php.net/manual/en/ref.session.php for explanation
-        if (!self::$_unitTestEnabled && defined('SID')) {
+        if (!self::$_unitTestEnabled && defined('SID') && SID !== '') {
             /** @see Zend_Session_Exception */
             // require_once 'Zend/Session/Exception.php';
             throw new Zend_Session_Exception('session has already been started by session.auto-start or session_start()');
@@ -662,7 +662,7 @@ class Zend_Session extends Zend_Session_Abstract
      */
     public static function setId($id)
     {
-        if (!self::$_unitTestEnabled && defined('SID')) {
+        if (!self::$_unitTestEnabled && defined('SID') && SID !== '') {
             /** @see Zend_Session_Exception */
             // require_once 'Zend/Session/Exception.php';
             throw new Zend_Session_Exception('The session has already been started.  The session id must be set first.');


falkenhawk:
@fredericgboutin-yapla could you try to add a test case to https://github.com/zf1s/zf1/blob/master/tests/Zend/Session/SessionTest.php covering the issue, please?

fredericgboutin-yapla:
I really hoped there would be some tests already covering my fix 😅 but yes sure.

falkenhawk:
> I really hoped there would be some tests already covering my fix 😅 but yes sure.

Nothing broke in existing tests with your change that's true, but a test should be added to cover the bug - which would fail without the fix and would pass after the fix.

packages/zend-session/library/Zend/Session.php
```diff
@@ -459,7 +459,7 @@ public static function start($options = false)
         }
 
         // See http://www.php.net/manual/en/ref.session.php for explanation
-        if (!self::$_unitTestEnabled && defined('SID')) {
+        if (!self::$_unitTestEnabled && defined('SID') && SID !== '') {
```
fredericgboutin-yapla, line 462: this condition is adding to the difficulty to properly unit-test

fredericgboutin-yapla:
there's already a test named `testInvalidPreexistingSessionIdDoesNotPreventRegenerationOfSid` that is playing with the mecanism I fixed.

Looking at it, I guess the problem is that it does not let PHP figure the `session_id` by itself through the `PHPSESSID` cookie.

Since I cannot call `setCookie` in this context, I tried replacing the `setcookie('PHPSESSID', 'xxx');` with `$_COOKIE['PHPSESSID'] = 'xxx';` and I think I reproduce the issue with the assert returning `Failed asserting that '' matches PCRE pattern "/^[0-9a-v]+$/".` 

Now let's try to put all this together...

fredericgboutin-yapla:
Hum... I think I got a bit out of my way in my last comment.

The current test is alright. The problem is really when,

1. there's no `session_id` already defined - in "real life" this happens on a brand new user, w/o any cookies
2. After skipping ID check, Zend starts a new session, PHP gets its `session_id` from the cookie and then sets `SID` to an empty string - see https://github.com/php/php-src/blob/PHP-7.0.33/ext/session/session.c#L1559
3. Zend is happy a session was created (and stored) despite having an ID that did NOT pass the check requirement since it was empty prior to calling `session_start`
4. the next time Zend is asked to "start" a session, check validation fails, it then "schedule" a `regenerateId` which does not happen because it bails out early when a session is already started.

So what do we do from here? Well... I guess we have to make the `getId()` and `_checkId()` statements at the beginning of the `start` function actually work when no session_id is currently defined.

Now about the `SID` constant...
The more I read about session management, the more I realize how old Zendframework 1 is since it doesn't support session strict mode - https://www.php.net/manual/en/session.configuration.php#ini.session.use-strict-mode - nor the session ID validation in session handlers.

I now really doubt the original purpose of Zend when using the `SID` constant to determine when a "session has already been started". Because, normally / by default, session management is done "using only cookies" - https://www.php.net/manual/en/session.configuration.php#ini.session.use-only-cookies - and not through URL parameters anymore (for security reasons). So it means, looking at the code of PHP 7.0.33, that the `SID` constant will either not exist OR always being set an empty string.

Long story short, we should most probably drop `SID` constant usage and simply rely on `session_status()` to know if a session was already started or not - https://www.php.net/manual/en/function.session-status.php

😩 I'm growing tired I think.

fredericgboutin-yapla:
We are successfully using a workaround like this,

```php
$sessionId = $_COOKIE['PHPSESSID'] ?? false;
if ($sessionId) {
	session_id($sessionId);
}
Zend_Session::start(...);
```

So maybe we could modify `getId` like,

```php
    public static function getId()
    {
		$id = session_id();
		
		if ($id === '' && isset($_COOKIE[session_name()])) {
			return $_COOKIE[session_name()];
		}

        return $id;
    }
```

And NOT forcing a re-generation when creating a new session with an invalid ID, 

```php
// Force a regenerate after session is started
if (self::$_sessionStarted) {
    self::$_regenerateIdState = -1;
}
```

I seems to work really well.

Now to test this. 🤔  We must NOT call `session_id(...)` at all and rely exclusively on cookie.

PR #178, opened 2023-12-17
update zend-view deps according to usage (closes #176)

### LINKED ISSUE #176: Zend_Locale doesn't seem to be a necessary dep of Zend_View
`Zend_Locale` is listed as a dep of `Zend_View`, but that only applies when using either the `Translate` or `Currency` view helpers.

Maybe because of this, it would be ok to move it to `suggest`? It sound unlikely that projects which need these helpers don't install either `Zend_Locale` directly or `Zend_Tranlate` (which it is a dep of).

(FWIW: I'm trying to migrate a project of of ZF1 on dep at a time and these inter-dependencies make it a bit of a challenge)

smhg:
I've listed everything used in view helpers under `suggest`.

That means a lot of additions, but also that `Zend_Controller` and `Zend_Locale` are no longer required dependencies. I can't imagine both are not installed explicitly when needed, but feel free to disagree of course.

### DIFF
--- a/packages/zend-view/composer.json
+++ b/packages/zend-view/composer.json
@@ -7,10 +7,8 @@
     "require": {
         "php": ">=5.3.3",
         "ext-reflection": "*",
-        "zf1s/zend-controller": "^1.15.2",
         "zf1s/zend-exception": "^1.15.2",
         "zf1s/zend-loader": "^1.15.2",
-        "zf1s/zend-locale": "^1.15.2",
         "zf1s/zend-registry": "^1.15.2"
     },
     "autoload": {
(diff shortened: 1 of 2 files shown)


falkenhawk:
@smhg maybe `zend-controller` was listed in required dependencies, because it's tightly coupled with `Zend_View_Helper_Action` which tends to be used a lot? 🤔
But on the other hand, to make it in line with other view helpers, all dependencies used by them should be optional 🤔 And that is what you did. The `action` helper doesn't deserve a special treatment, does it?

smhg:
FWIW: in my project Zend_View is used for email template rendering, independent of Zend_Controller. It is probably not common usage.

But I do think coupling should be on the loose side. Because in the end, projects should likely be refactoring away from ZF1 and thus it is useful to require only the absolute minimum rather than the safe maximum.

And, when using zf1s as a zf1 replacement, what you are looking for includes 'less coupling' probably?

Also, specifically in the Zend_Controller case, I think it must be quite uncommon to only specify Zend_View in your project and count on it including Zend_Controller, no?

Feel free to disagree of course.

falkenhawk (approved):
👍

PR #179, opened 2023-12-18
update zend-validate deps according to usage

smhg:
This also brings `Zend_Validate` deps in line with usage: everything used in the base classes now is a required dep, while everything in the different validators is listed as a suggested dep.


falkenhawk (approved):
thanks a lot @smhg !

PR #186, opened 2024-01-09
[zend-auth] update zend-auth deps according to usage

smhg:
@falkenhawk just some missing suggested deps for Zend_Auth.


falkenhawk (approved):
Thanks!

PR #187, opened 2024-01-09
[zend-config] update zend-config deps according to usage

smhg:
@falkenhawk brings Zend_Config deps up to date, ~~detatching Zend_Xml~~.

### DIFF
--- a/packages/zend-config/composer.json
+++ b/packages/zend-config/composer.json
@@ -15,7 +15,8 @@
         }
     },
     "suggest": {
-        "ext-simplexml": "Used in special situations or with special adapters (Zend_Config_Xml)"
+        "ext-simplexml": "Used in special situations or with special adapters (Zend_Config_Xml)",
+        "zf1s/zend-json": "Used in special situations or with special adapters"
     },
     "replace": {
         "zf1/zend-config": "^1.12"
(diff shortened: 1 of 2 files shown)


falkenhawk (changes requested):
Thanks for preparing the PR!

I need to reject it, it's very subjective but we use xml config format across all our projects. Removing zend-xml from package dependencies will make it crash after update if we don't remember to add zend-xml to project dependencies, as Zend_Config uses Zend_Xml_Security right in the constructor, so it's tightly coupled.

I hope you'll understand @smhg - could you perhaps modify the PR, leaving zend-xml where it was, and only adding zend-json to `suggest`, please?

smhg:
@falkenhawk Thank you for all your work!

I totally understand if you have to be pragmatic and can't justify this change in your own projects.
In that case, I think we need to add Zend_Json, Zend_Xml and Zend_Yaml as required deps.

As a last attempt: in my project I only use Zend_Config_Ini. So I have no use for these dependencies. In my long-term attempt to get rid of ZF1 components, it sounds great to decouple them.

But again, I am fine with your decission. Shall I add the deps?

glensc:
7d6b78fcd9f4ac43abaf8d1348d9c0a059704b50 could be extracted to a separate PR as that is safe to be merged

falkenhawk (approved):

PR #188, opened 2024-01-09
[zend-controller] update zend-controller deps according to usage

smhg:
@falkenhawk this brings Zend_Controller deps up to date.
There is a lot of room for interpretation here... I've taken the least intrusive path.
But in theory Zend_View and Zend_Filter aren't necessarily required dependencies I think.


falkenhawk (approved):
👍

PR #192, opened 2024-01-09
[zend-rest] update zend-rest deps according to usage

smhg:
@falkenhawk this is the last component (at least for now). Thanks!

### DIFF
--- a/packages/zend-rest/composer.json
+++ b/packages/zend-rest/composer.json
@@ -24,7 +24,6 @@
         }
     },
     "suggest": {
-        "zf1s/zend-config": "Used in special situations or with special adapters"
     },
     "replace": {
         "zf1/zend-rest": "^1.12"
--- a/packages/zend-rest/composer.json
+++ b/packages/zend-rest/composer.json
@@ -10,7 +10,9 @@
         "ext-dom": "*",
         "ext-reflection": "*",
         "ext-simplexml": "*",
+        "zf1s/zend-controller": "^1.15.5",
         "zf1s/zend-exception": "^1.15.5",
+        "zf1s/zend-http": "^1.15.5",
         "zf1s/zend-server": "^1.15.5",
         "zf1s/zend-service": "^1.15.5",
         "zf1s/zend-uri": "^1.15.5",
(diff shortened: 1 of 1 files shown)


packages/zend-rest/composer.json
```diff
@@ -10,7 +10,9 @@
         "ext-dom": "*",
         "ext-reflection": "*",
         "ext-simplexml": "*",
+        "zf1s/zend-controller": "^1.15.3",
```
falkenhawk, line 13: could we keep `zend-controller` in `suggest`, please? It's required only by `Zend_Rest_Controller` and `Zend_Rest_Route`. So it's optional when only client/server classes are used.

packages/zend-rest/composer.json
```diff
@@ -21,6 +23,9 @@
             "Zend_Rest": "library/"
         }
     },
+    "suggest": {
+        "zf1s/zend-config": "Used in special situations or with special adapters"
```
falkenhawk, line 27: `Zend_Config` is not used at all - can be removed from `suggest`. Type hinting an instance of `Zend_Config` in `Zend_Rest_Route::getInstance()` does not make the class/package itself required.

falkenhawk (changes requested):

packages/zend-rest/composer.json
smhg: Just so I understand: the type hinting, if not installed, won't lead to an error?
Or you mean type hinting-only use is not enough to make it required?

packages/zend-rest/composer.json
smhg: This was indeed a close call. I myself only use Zend_Rest_Controller and Zend_Rest_Route (and not the client/server classes). Both uses are indeed completely separate. `suggest` sounds better.

packages/zend-rest/composer.json
falkenhawk: Both are correct

packages/zend-rest/composer.json
smhg: @falkenhawk Can all other deps also be moved to suggest? With the exception of `Zend_Exception`. That makes sense when using only the MVC classes (Zend_Rest_Controller and Zend_Rest_Route).

smhg: @falkenhawk I'm not sure what you meant with the eyes emoji :) I won't move the others in this PR for now.

smhg:
@falkenhawk I think this should be fine to merge now too?

falkenhawk (approved):
Thank you @smhg !

PR #194, opened 2024-02-13
[zend-form] fix instantiating filters with options via `addFilter()` in php 8.0+

marcing:
PHP 8.0+ changes behaviour of ReflectionClass::newInstanceArgs, which now assumes that an associative array is an array of named parameters, trying to match array keys to names of arguments in the constructor. This leads to errors in Zend_Form_Element::_loadFilter.

```php
$this->addFilter('Boolean', array('type' => Zend_Filter_Boolean::PHP + Zend_Filter_Boolean::FALSE_STRING));
```

```
zend-form\library\Zend\Form\Element.php:2117
Error: Unknown named parameter $type
```

Zend_Form_Element::_loadFilter has to be altered in the same manner as _loadValidator to call newInstanceArgs only or numerical arrays and call newInstance on associative arrays.

### DIFF
--- a/tests/Zend/Form/ElementTest.php
+++ b/tests/Zend/Form/ElementTest.php
@@ -1226,15 +1226,57 @@ class Zend_Form_ElementTest extends PHPUnit_Framework_TestCase
         $this->assertTrue($filter->allowWhiteSpace);
     }
 
-    public function testShouldUseFilterConstructorOptionsAsPassedToAddFilter()
+    public function testCanPassFirstAndSecondArgument()
     {
-        $this->element->addFilter('HtmlEntities', array(array('quotestyle' => ENT_QUOTES, 'charset' => 'UTF-8')));
+        try {
+            // public function __construct($searchSeparator = ' ', $replacementSeparator = '-')
+            $this->element->addFilter('Word_SeparatorToSeparator', array('-', '_'));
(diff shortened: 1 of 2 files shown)


tests/Zend/Form/ElementTest.php
```diff
@@ -1225,6 +1225,30 @@ public function testOptionsAreCastToArrayWhenAddingFilter()
...
+        }
+        $filter = $this->element->getFilter('Boolean');
+        $this->assertTrue($filter instanceof Zend_Filter_Boolean);
+        $this->assertSame($filter->getType(), Zend_Filter_Boolean::PHP + Zend_Filter_Boolean::FALSE_STRING);
+    }
+    
+    public function testFilterOptionsAsArray()
+    {
+        try {
+            $this->element->addFilter('HtmlEntities', array(array('encoding' => 'MacRoman')));
```
falkenhawk, line 1244: I think this test case is redundant - nested assoc array in an array is already tested in the next test right below it - `testShouldUseFilterConstructorOptionsAsPassedToAddFilter()`

A test which checks if a filter can be instantiated with an array of arguments via a numerical array.

For instance, `addFilter('SeparatorToSeparator', array('-', '_'))`

falkenhawk (approved):
Thanks @marcing ! 💚

PR #199, opened 2024-07-26
Set php 5.4 as minimum requirement

sreichel:
I'd like to sync some changes from ZF1Future, but traditional array sytax makes it hard to diff changes.

PR updates minimum php requirement to 5.4 and changed to short array syntax.


sreichel:
Re-opend.

Do we really need 5.3-support anymore?

glensc:
You can use rector or php-cs-fixer to convert the array to long syntax.

glensc:
also: never put automated changes (formatting changes) with other changes to the same commit!

format commits could be put to git blame ignore file (`.git-blame-ignore-revs file`) which github supports now:
- https://github.com/eventum/eventum/pull/1552

sreichel:
Hey, thanks.

Do you mean the changes to phpcs config and composer.json?

The idea for this PR was to port some change from ZF1Future.

Problem, I have a lot of noise in diff, because of traditional array syntax.

Of course I could convert it to short syntax, apply the changes, but then I have to revert it to long syntax befor commit my changes.

I'd like to switch from ZF1Future to ZF1s, b/c OpenMage only uses a few Zend libraries.

Therefore I have to port some fixes.

falkenhawk:
@sreichel thanks for your efforts here, and your intention to port fixes from zf1-future, that might be beneficial, but as @glensc try to take the opposite approach - checkout zf1-future, use whatever tool to convert array syntax to the long one, and port fixes from there. Or simply resolve conflicts while cherrypicking individual commits from zf1-future.

We won't be modernizing the codebase across the whole projects only for the sake of it, that misses the point here.

PR #202, opened 2025-01-28
[zend-db] update $_schema type hint from array to string

Marrikulus:
Update $_schema type hint from array to string.
It's only ever used like a string and never as an array.

I discovered this while using phpstan on my codebase.
I have over 500 tables that inherited from this class and all of them define the schema property.


falkenhawk (approved):
Thanks!

PR #203, opened 2025-04-10
Fix: Passing null is deprecated in `Zend_Mail::_filterEmail()`

sreichel:
- closes https://github.com/OpenMage/magento-lts/issues/3706


falkenhawk:
Hi @sreichel, thanks for the PR!

Is there an issue though? Why it should be allowed to pass in NULL as a "from", "cc", or "bcc" email and have it silently converted to an empty string?

Maybe that the deprecation error is now triggered, it could be actually helpful to find the real cause of the issue in your integration?

sreichel:
php silently converted it to string.

PR #207, opened 2025-04-17
PHP 8.4: Implicitly nullable parameter declarations deprecated

marcing:
PHP 8.4: Implicitly nullable parameter declarations deprecated

https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated

Fixes error:
Implicitly marking parameter ? as nullable is deprecated, the explicit nullable type must be used instead

Fix using Zf1s\Compat\Types


marcing:
@dmnc No actually in the end I managed to make it work on all versions 5.6-8.4. I had to introduce a helper function though, because there is no syntax that would work in both 5.6 and 8.4.

falkenhawk:
superseded by #216 
Thank you @marcing for your huge efforts here!

PR #230, opened 2026-04-08
catch Throwable in dispatch and server error boundaries

### LINKED ISSUE #201: Front controller won't catch exceptions of type Error
## Context

In recent versions of PHP, several so called errors migrated to be thrown as exceptions instead. The wording is confusing here, but the new class they created is named [`Error`](https://www.php.net/manual/en/class.error.php) and it does NOT inherit from [`Exception`](https://www.php.net/manual/en/class.exception.php) but directly from `Throwable`.

## What does that mean?

Let's take an example, a division by zero.

In PHP 7, dividing by zero would generate an error of type `E_Warning` and continue its execution.

In PHP 8, dividing by zero leads to a [`DivisionByZeroError`](https://www.php.net/manual/en/class.divisionbyzeroerror.php) exception being thrown.

Let's imagine that division by zero happened when dispatching / rendering a page, for example in the `Action` of a `Controller`. Normally, in that context, any `Exception` would "bubble up" through the `Front` and the `Standard` controllers `dispatch()`methods and be handled properly by Zend - see https://github.com/zf1s/zf1/blob/7d4b8fe7848bacd0836cc7fa5f987c7b1f0fbb2e/packages/zend-controller/library/Zend/Controller/Front.php#L980C8-L980C33

The problem is, since a `DivisionByZeroError` is NOT an `Exception` it doesn't get catch by Zend and it ends up as a global exception.

## What's the problem

Well, `Errors` (and any `Throwable`)  are ending up as global unhandled exceptions, with no graceful "bailout" or anything and this is something that was introduced in PHP 7 but that is getting more frequent in PHP 8 and would probably continue like that in future PHP versions as they are laying more on exceptions more than errors.

## Proposition

I suppose we should consider revisiting those `try/catch` in the controllers involved in dispatching, and change `Exception` to `Throwable`, or at the very least add the `Error` alongside `Exception` handling. 

The minimal would be to adjust both the Front and the Standard controllers.

What do you think ?

falkenhawk:
closes #201

Since PHP 7, `Error` classes (`TypeError`, `DivisionByZeroError`, etc.) extend `Throwable` but not `Exception`. The framework's `catch (Exception)` blocks in dispatch chains and server handlers let these slip through as unhandled crashes.

Mainly affects `zend-controller` (front controller, dispatcher, plugin broker). The same pattern is also fixed in server components (`zend-rest`, `zend-soap`, `zend-json`, `zend-xmlrpc`) and `zend-tool`.

Now that PHP 7.1 is the minimum (#218), all error boundary catch blocks are widened to `catch (Throwable)`.

#### `setException()` compatibility

`Zend_Controller_Response_Abstract::setException()` and `Zend_Tool_Framework_Client_Response::setException()` type-hint `Exception`. To avoid a BC break in the method signature, non-Exception throwables are wrapped in the appropriate framework exception class before being passed, with the original preserved as `$previous`.

Wrapping means that e.g. a `DivisionByZeroError` will appear in the response as `Zend_Controller_Exception`, so `hasExceptionOfType('DivisionByZeroError')` won't match - but the original is accessible via `getPrevious()`. This is a non-issue in practice: nobody can be checking for `Error` subtypes in the response because they were never caught before. Any existing `hasExceptionOfType()` usage checks for `Exception` subclasses, which still pass through unwrapped.

If you need to handle a specific error type, catch it directly in your controller action before it reaches the dispatch boundary.

#### server `fault()` methods

`instanceof Exception` checks in `Zend_Rest_Server::fault()`, `Zend_Soap_Server::fault()`, and `Zend_XmlRpc_Server::fault()` are updated to `instanceof Throwable` so error details (message, code) are properly extracted instead of falling through to generic "Unknown error" messages.

### DIFF
--- a/packages/zend-controller/library/Zend/Controller/Front.php
+++ b/packages/zend-controller/library/Zend/Controller/Front.php
@@ -287,7 +287,7 @@ class Zend_Controller_Front
     {
         try{
             $dir = new DirectoryIterator($path);
-        } catch(Exception $e) {
+        } catch(Throwable $e) {
             // require_once 'Zend/Controller/Exception.php';
             throw new Zend_Controller_Exception("Directory $path not readable", 0, $e);
         }
@@ -909,11 +909,13 @@ class Zend_Controller_Front
 
             try {
                 $router->route($this->_request);
-            }  catch (Exception $e) {
+            }  catch (Throwable $e) {
                 if ($this->throwExceptions()) {
                     throw $e;
                 }
-
+                if (!$e instanceof Exception) {
+                    $e = new Zend_Controller_Exception($e->getMessage(), $e->getCode(), $e);
+                }
                 $this->_response->setException($e);
             }
 
@@ -952,9 +954,12 @@ class Zend_Controller_Front
                  */
                 try {
                     $dispatcher->dispatch($this->_request, $this->_response);
-                } catch (Exception $e) {
+                } catch (Throwable $e) {
                     if ($this->throwExceptions()) {
                         throw $e;
+                    }
+                    if (!$e instanceof Exception) {
+                        $e = new Zend_Controller_Exception($e->getMessage(), $e->getCode(), $e);
                     }
                     $this->_response->setException($e);
                 }
@@ -964,11 +969,13 @@ class Zend_Controller_Front
                  */
                 $this->_plugins->postDispatch($this->_request);
             } while (!$this->_request->isDispatched());
-        } catch (Exception $e) {
+        } catch (Throwable $e) {
             if ($this->throwExceptions()) {
                 throw $e;
             }
-
+            if (!$e instanceof Exception) {
+                $e = new Zend_Controller_Exception($e->getMessage(), $e->getCode(), $e);
+            }
             $this->_response->setException($e);
         }
 
@@ -977,11 +984,13 @@ class Zend_Controller_Front
          */
         try {
             $this->_plugins->dispatchLoopShutdown();
-        } catch (Exception $e) {
+        } catch (Throwable $e) {
             if ($this->throwExceptions()) {
                 throw $e;
             }
-
+            if (!$e instanceof Exception) {
+                $e = new Zend_Controller_Exception($e->getMessage(), $e->getCode(), $e);
+            }
             $this->_response->setException($e);
         }
 
--- a/packages/zend-controller/library/Zend/Controller/Dispatcher/Standard.php
+++ b/packages/zend-controller/library/Zend/Controller/Dispatcher/Standard.php
@@ -306,7 +306,7 @@ class Zend_Controller_Dispatcher_Standard extends Zend_Controller_Dispatcher_Abs
 
         try {
             $controller->dispatch($action);
-        } catch (Exception $e) {
+        } catch (Throwable $e) {
             // Clean output buffer on error
             $curObLevel = ob_get_level();
             if ($curObLevel > $obLevel) {
--- a/packages/zend-controller/library/Zend/Controller/Plugin/Broker.php
+++ b/packages/zend-controller/library/Zend/Controller/Plugin/Broker.php
@@ -238,10 +238,13 @@ class Zend_Controller_Plugin_Broker extends Zend_Controller_Plugin_Abstract
         foreach ($this->_plugins as $plugin) {
             try {
                 $plugin->routeStartup($request);
-            } catch (Exception $e) {
+            } catch (Throwable $e) {
                 if (Zend_Controller_Front::getInstance()->throwExceptions()) {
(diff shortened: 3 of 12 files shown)


falkenhawk:
@fredericgboutin-yapla would you please take a look, if this is still relevant to you?

packages/zend-controller/library/Zend/Controller/Front.php
```diff
@@ -964,11 +969,13 @@ public function dispatch(?Zend_Controller_Request_Abstract $request = null, ?Zen
...
             } while (!$this->_request->isDispatched());
-        } catch (Exception $e) {
+        } catch (Throwable $e) {
             if ($this->throwExceptions()) {
                 throw $e;
             }
-
+            if (!$e instanceof Exception) {
+                $e = new Zend_Controller_Exception($e->getMessage(), $e->getCode(), $e);
+            }
```
partikus, line 978: should we decorate it before forwarding?

```suggestion
            if (!$e instanceof Exception) {
                $e = new Zend_Controller_Exception($e->getMessage(), $e->getCode(), $e);
            }
            if ($this->throwExceptions()) {
                throw $e;
            }
```

packages/zend-controller/library/Zend/Controller/Front.php
```diff
@@ -909,11 +909,13 @@ public function dispatch(?Zend_Controller_Request_Abstract $request = null, ?Zen
...
                 $router->route($this->_request);
-            }  catch (Exception $e) {
+            }  catch (Throwable $e) {
                 if ($this->throwExceptions()) {
                     throw $e;
                 }
-
+                if (!$e instanceof Exception) {
+                    $e = new Zend_Controller_Exception($e->getMessage(), $e->getCode(), $e);
+                }
```
partikus, line 918: Should we decorate before forwarding?

```suggestion
                if (!$e instanceof Exception) {
                    $e = new Zend_Controller_Exception($e->getMessage(), $e->getCode(), $e);
                }
                if ($this->throwExceptions()) {
                    throw $e;
                }
               
```

packages/zend-controller/library/Zend/Controller/Front.php
```diff
@@ -977,11 +984,13 @@ public function dispatch(?Zend_Controller_Request_Abstract $request = null, ?Zen
          */
         try {
             $this->_plugins->dispatchLoopShutdown();
-        } catch (Exception $e) {
+        } catch (Throwable $e) {
             if ($this->throwExceptions()) {
                 throw $e;
             }
-
+            if (!$e instanceof Exception) {
```
partikus, line 991: idem

partikus (commented):
it looks ok, just few thoughts

fredericgboutin-yapla:
Looks good to me once we observe the decorators when applicable, as suggested.

packages/zend-controller/library/Zend/Controller/Front.php
falkenhawk: should really `Error`-s be treated that way, wrapped with `Zend_Controller_Exception`, even when `throwExceptions()` is true? I'd rather expect them to be re-thrown as is 🤔

falkenhawk: heh but on the other hand, in `Zend_Controller_Plugin_Broker`, i.a. https://github.com/zf1s/zf1/pull/230/changes#diff-51643978e1b538318ad62e0bfe12d049ede90a15231adb3574018f76c9812821L243
all exceptions caught are wrapped with `Zend_Controller_Exception`, when `throwExceptions()` is true. 🤡

falkenhawk: i delegated the investigation on this to claude and here are the results:

---

Should Front.php also wrap before rethrowing when `throwExceptions=true`?

No, and here's why. The two behaviors are intentionally different in the original ZF1 code:

| | `throwExceptions=true` | `throwExceptions=false` |
|---|---|---|
| **Broker** | wraps ALL throwables in `Zend_Controller_Exception` | wraps only non-Exception, stores in response |
| **Front** | rethrows raw | wraps only non-Exception, stores in response |

The Broker wraps even regular `Exception` subclasses when rethrowing - it's a normalization layer for plugin code. Front.php never wraps on rethrow - it preserves whatever came through. This is the original design, predating our changes.

If we started wrapping only `Error` types (but not `Exception` types) in Front's `throwExceptions` path, that would be inconsistent - different wrapping rules depending on the exception class hierarchy.

In practice, Errors from plugins never reach Front raw - the Broker wraps them first. The only raw Errors that hit Front directly come from `$router->route()` or `$dispatcher->dispatch()` - framework internals where preserving the original type is more useful.

Added tests in 9724f4d1 covering both paths - plugin errors (wrapped by Broker) and controller action errors (raw `TypeError` rethrown by Front).

PR #232, opened 2026-07-02
Fix Zend_Db::FETCH_* modifier-flag drift from PDO on PHP 8.5

### LINKED ISSUE #231: Zend_Db::FETCH_* modifier flags drift from PDO::FETCH_* on PHP 8.5 → ValueError
# `Zend_Db::FETCH_GROUP`/`FETCH_UNIQUE`/`FETCH_CLASSTYPE`/`FETCH_SERIALIZE` drift from `PDO::FETCH_*` on PHP 8.5 → `ValueError` in the PDO adapter

## Summary

On **PHP 8.5**, `Zend_Db_Statement_Pdo::fetchAll()` / `fetch()` / `setFetchMode()`
throw `ValueError: ... must be a bitmask of PDO::FETCH_* constants` whenever a
fetch **modifier flag** (`FETCH_GROUP`, `FETCH_UNIQUE`, `FETCH_CLASSTYPE`,
`FETCH_SERIALIZE`) is used — on its own or combined, e.g. `FETCH_GROUP | FETCH_COLUMN`.

## Environment

- zf1s/zf1 **master** and latest tag **1.16.0** (both affected)
- PHP **8.5** (reproduced on 8.5.7)
- PDO adapter (Pgsql / MySQL)

## Root cause

`Zend/Db.php` freezes the PDO fetch-mode values as integer literals. Its own
docblock shows they were captured once by iterating `constant("PDO::$name")`:

```php
const FETCH_GROUP     = 65536;
const FETCH_UNIQUE    = 196608;
const FETCH_CLASSTYPE = 262144;
const FETCH_SERIALIZE = 524288;
```

`Zend_Db_Statement_Pdo` forwards the fetch mode **verbatim** to the native
statement, relying on `Zend_Db::FETCH_* === PDO::FETCH_*`:

```php
// Zend/Db/Statement/Pdo.php
return $this->_stmt->fetchAll($style);   // $style is the raw Zend_Db int
```

PHP 8.5 renumbered the PDO fetch **modifier flags** — observed: `FETCH_GROUP`
`65536 → 32`, `FETCH_UNIQUE` `196608 → 64` (and `FETCH_CLASSTYPE` /
`FETCH_SERIALIZE` likewise). So `Zend_Db::FETCH_GROUP` (65536) is no longer a
valid PDO fetch bitmask and PDO rejects it with a `ValueError`, which the
adapter already catches (added in #85) and re-wraps as
`Zend_Db_Statement_Exception` — but the underlying value drift is unaddressed.

**Base modes are unaffected** (`FETCH_ASSOC`=2, `FETCH_NUM`=3, `FETCH_BOTH`=4,
`FETCH_OBJ`=5, `FETCH_COLUMN`=7, …) — their integer values are unchanged on
8.5. Only the high-bit modifier flags drifted.

The recent PHP 8.5 compatibility PRs (#225–#228) fixed deprecations but did
**not** touch `Zend/Db.php` or `Zend/Db/Statement/Pdo.php`, so this specific
issue remains open on `master`.

## Reproduction

```php
$db->query('SELECT k, v FROM t')->fetchAll(Zend_Db::FETCH_GROUP);
// PHP 8.5  → ValueError → Zend_Db_Statement_Exception
// PHP ≤ 8.4 → works
```

## Suggested fix

Translate the fetch mode to native PDO values **at the PDO adapter boundary**
(where `ext-pdo` is guaranteed) rather than relying on numeric equality. This
keeps the public `Zend_Db::FETCH_*` literals stable — external code compares
against them, and redefining them as `const FETCH_GROUP = PDO::FETCH_GROUP;`
would fatal at class load on non-PDO adapters (Mysqli / Oracle / Db2 / Sqlsrv)
where `ext-pdo` may be absent (which is presumably why they were frozen).

One subtlety worth flagging: the legacy literals overlap in their bit patterns
(`FETCH_UNIQUE` = `0x30000` embeds the `FETCH_GROUP` bit `0x10000`), so a naive
per-bit remap leaks `FETCH_GROUP` into a `FETCH_UNIQUE`-only request. The remap
must strip the legacy flag bits and re-add PDO's runtime values widest-flag-first.

I have a PR ready with this fix in `Zend_Db_Statement_Pdo` plus a DB-less
regression test (identity on PHP < 8.5, correct remap on 8.5). Happy to open it
if this direction looks right.

kodown1k:
## What

Remap the `Zend_Db::FETCH_*` fetch-mode **modifier flags** to their native
`PDO::FETCH_*` values inside `Zend_Db_Statement_Pdo` before forwarding a fetch
mode to the underlying `PDOStatement`.

Fixes the `ValueError: ... must be a bitmask of PDO::FETCH_* constants` thrown
on **PHP 8.5** by `fetch()` / `fetchAll()` / `setFetchMode()` whenever
`FETCH_GROUP`, `FETCH_UNIQUE`, `FETCH_CLASSTYPE` or `FETCH_SERIALIZE` is used.

## Why

`Zend_Db::FETCH_*` are integer literals frozen from an old PDO snapshot
(`FETCH_GROUP = 65536`, `FETCH_UNIQUE = 196608`, …). PHP 8.5 renumbered the PDO
fetch modifier flags (`FETCH_GROUP` `65536 → 32`, `FETCH_UNIQUE` `196608 → 64`,
etc.), so the frozen literals no longer equal PDO's runtime constants and PDO
rejects them. Base fetch modes (`ASSOC`/`NUM`/`COLUMN`/…) are unchanged and
unaffected.

## Approach

Translate at the adapter boundary (`ext-pdo` guaranteed there) instead of
redefining the `Zend_Db::FETCH_*` constants — the latter would fatal at class
load on non-PDO adapters (Mysqli/Oracle/Db2/Sqlsrv) without `ext-pdo`, and would
change a public API surface that external code compares against.

The remap strips the legacy flag bits and re-adds PDO's runtime values
widest-flag-first, because the legacy literals overlap
(`FETCH_UNIQUE` `0x30000` embeds the `FETCH_GROUP` bit `0x10000`) — a naive
per-bit swap would leak `FETCH_GROUP` into a `FETCH_UNIQUE`-only request.

On **PHP < 8.5** every mapping is the identity, so this is a no-op there.

## Tests

`tests/Zend/Db/Statement/PdoFetchModeTest.php` — DB-less, exercises the
translation via reflection: base modes pass through, each modifier flag maps to
its `PDO::FETCH_*` equivalent, combined bitmasks (`FETCH_GROUP | FETCH_COLUMN`)
remap correctly, and the `FETCH_UNIQUE ⊃ FETCH_GROUP` overlap is guarded.
Meaningful on every version in the CI matrix (identity on ≤ 8.4, real remap on 8.5).

Fixes #231

### DIFF
--- a/packages/zend-db/library/Zend/Db/Statement/Pdo.php
+++ b/packages/zend-db/library/Zend/Db/Statement/Pdo.php
@@ -254,7 +254,7 @@ class Zend_Db_Statement_Pdo extends Zend_Db_Statement implements IteratorAggrega
             $cursor = PDO::FETCH_ORI_NEXT;
         }
         try {
-            return $this->_stmt->fetch($style, $cursor, $offset);
+            return $this->_stmt->fetch(self::_normalizeFetchMode($style), $cursor, $offset);
         } catch (ValueError $e) {
             // PHP 8.0+ throws ValueError on invalid arguments
             throw new Zend_Db_Statement_Exception($e->getMessage(), $e->getCode(), $e);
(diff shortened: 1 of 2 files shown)


falkenhawk (approved):
Thank you!

falkenhawk:
Released in [v1.16.1](https://github.com/zf1s/zf1/releases/tag/1.16.1).

PR #233, opened 2026-08-21
doc: fix phpdoc of Select::from

sd24pat:
_join already handles Zend_Db_Select but static analyzers complain because the phpdoc does not show this


falkenhawk (approved):
Thank you @sd24pat !

